PR #10, opened 2021-06-28
Add option to config the axios requests

claudiulataretu:
Signed-off-by: Claudiu Ion Lataretu <claudiu.lataretu@gmail.com>

### DIFF
--- a/src/apiProvider.ts
+++ b/src/apiProvider.ts
@@ -1,4 +1,4 @@
-import axios from "axios";
+import axios, { AxiosRequestConfig } from "axios";
 import { IApiProvider } from "./interface";
 import * as errors from "./errors";
 import { Logger } from "./logger";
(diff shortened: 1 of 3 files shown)


danielailie (approved):

src/apiProvider.ts
```diff
@@ -16,16 +16,19 @@ const JSONbig = require("json-bigint");
...
+  private config: AxiosRequestConfig;
 
   /**
    * Creates a new ApiProvider.
    * @param url the URL of the Elrond Api
    * @param timeout the timeout for any request-response, in milliseconds
+   * @param config axios request config options
    */
-  constructor(url: string, timeout?: number) {
+   constructor(url: string, timeout?: number, config?: AxiosRequestConfig) {
```
ccorcoveanu, line 27: Would it be a pain maybe if this won't be backwards compatible? The timeout param is in `AxiosRequestConfig`, I am thinking we should remove it from the constructor and use it in the newly config param.

claudiulataretu: It should not be an issue for the way I'm using it for service. The problem is that this erdjs module is used heavily for all the web applications and I don't know that might be the implications to not keep it backwards compatible.

danielailie (dismissed):

danielailie (approved):

ccorcoveanu (approved):

PR #11, opened 2021-06-29
Encryption component

ccorcoveanu:
Added new component for `EncryptedData`.

It uses the same key derivation mechanism as the account keyfile (in fact, that was refactored to use `EncryptedData` under the hood which is more general-purpose).

Link to component description:
https://docs.google.com/document/d/1Xxb39JJkf8B9UzzQZd3KeOzL6nomoOxOyxKpxDKRgiE/edit#


src/walletcore/userWallet.ts
```diff
@@ -85,14 +46,23 @@ export class UserWallet {
...
 
-    /**
-     * Will take about:
-     *  - 80-90 ms in Node.js, on a i3-8100 CPU @ 3.60GHz
-     *  - 350-360 ms in browser (Firefox), on a i3-8100 CPU @ 3.60GHz
-     */
-    private static generateDerivedKey(password: Buffer, salt: Buffer, kdfparams: ScryptKeyDerivationParams): Buffer {
-        const derivedKey = scryptsy(password, salt, kdfparams.n, kdfparams.r, kdfparams.p, kdfparams.dklen);
-        return derivedKey;
+    static ecFromJSON(keyfileObject: any): EncryptedData {
```
andreibancioiu, line 49: What does the prefix `ec` mean?

src/crypto/encrypt.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { assert } from "chai";
```
andreibancioiu, line 1: 👍 for this test.

src/crypto/decryptor.ts
```diff
@@ -0,0 +1,27 @@
...
+    const computedMAC = crypto.createHmac(DigestAlgorithm, derivedKeySecondHalf).update(ciphertext).digest();
+    const actualMAC = data.mac;
+
+    if (computedMAC.toString("hex") !== actualMAC) {
+      throw new errors.ErrWallet("MAC mismatch, possibly wrong password");
+    }
+
+    const decipher = crypto.createDecipheriv(data.cipher, derivedKeyFirstHalf, iv);
+
+    return Buffer.concat([decipher.update(ciphertext), decipher.final()]);
```
andreibancioiu, line 25: So, the padding is still done in `userWallet.ts`.

I can confirm there's no change in logic - only refactoring ✔️

src/crypto/encryptor.ts
```diff
@@ -0,0 +1,30 @@
+import crypto from "crypto";
+import { Randomness } from "./randomness";
+import { ScryptKeyDerivationParams } from "./derivationParams";
+import { CipherAlgorithm, DigestAlgorithm, Version, KeyDerivationFunction } from "./constants";
+import {EncryptedData} from "./encryptedData";
+
+export class Encryptor {
+  public static encrypt(data: Buffer, password: string, randomness: Randomness = new Randomness()): EncryptedData {
+    const kdParams = new ScryptKeyDerivationParams();
```
andreibancioiu, line 9: I confirm the login here is the one from the previous constructor of `UserWallet` ✔️

andreibancioiu (approved):
👍

src/walletcore/userWallet.ts
ccorcoveanu: It was a typo, should have been `ed` for "Encrypted Data"

andreibancioiu (approved):

PR #17, opened 2021-07-06
Provider sign APIs

ccorcoveanu:
Implemented APIs for transaction signing
+ Some small refactoring in the wallet connect component

### DIFF
--- a/src/dapp/walletProvider.ts
+++ b/src/dapp/walletProvider.ts
@@ -241,13 +241,7 @@ export class WalletProvider implements IDappProvider {
         }
 
         return new Promise<Transaction>((resolve, reject) => {
-            let plainTransaction = transaction.toPlainObject();
-
-            // We adjust the fields, in order to make them compatible with what the wallet expected
-            plainTransaction["data"] = transaction.getData().valueOf().toString();
-            plainTransaction["value"] = transaction.getValue().toString();
-            plainTransaction["gasPrice"] = transaction.getGasPrice().valueOf();
-            plainTransaction["gasLimit"] = transaction.getGasLimit().valueOf();
+            let plainTransaction = WalletProvider.prepareWalletTransaction(transaction);
             console.log("postMessage", DAPP_MESSAGE_SEND_TRANSACTION_URL, plainTransaction);
 
             contentWindow.postMessage({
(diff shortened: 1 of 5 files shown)


radumojic (dismissed):

src/dapp/walletProvider.ts
```diff
@@ -285,6 +287,70 @@ export class WalletProvider implements IDappProvider {
...
+            throw new Error("Wallet provider is not initialised, call init() first");
+        }
+
+        const {contentWindow} = this.mainFrame;
+        if (!contentWindow) {
+            throw new Error("Wallet provider is not initialised, call init() first");
+        }
+
+        return new Promise<Transaction>((resolve, reject) => {
+            let plainTransaction = transaction.toPlainObject();
```
danielailie, line 307: can you extract this into a different method, something like prepareWalletTransaction or something similar

src/dapp/walletProvider.ts
```diff
@@ -285,6 +287,70 @@ export class WalletProvider implements IDappProvider {
...
+
+            contentWindow.postMessage({
+                type: DAPP_MESSAGE_SIGN_TRANSACTION_URL,
+                data: {
+                    transaction: plainTransaction
+                }
+            }, this.walletUrl);
+
+            const timeout = setTimeout(_ => reject('sign transaction url not responding'), 5000);
+            const signTransactionUrl = (ev: IDappMessageEvent) => {
```
danielailie, line 325: also here, can you extract into getTransactionUrl

src/dapp/walletProvider.ts
ccorcoveanu: done

src/dapp/walletProvider.ts
ccorcoveanu: I don't get why - this is already a function. It's encapsulated in the class method because it does not have a point to be an actual method itself. It follows the same pattern in all functions :)

danielailie: it was for readability, it easier to follow the code when there are smaller methods

danielailie (approved):

radumojic (approved):

PR #28, opened 2021-07-15
Integrate message signing

ccorcoveanu:
Code from this PR is in great part integrated thanks to @alexandru-g who opened this PR in the old repo:
https://github.com/ElrondNetwork/elrond-sdk/pull/215/files

There are some modifications though, mostly related to the architecture following the same pattern as our transaction signers. Also all DAPP providers (except HWWallet) throw exceptions on `signMessage` because the prepended message should be **enforced** by the provider itself.


src/interface.ts
```diff
@@ -124,6 +129,20 @@ export interface ISignable {
   applySignature(signature: Signature, signedBy: Address): void;
 }
 
+/**
+ * Interface that defines a signed and verifieble object
```
bogdan-rosianu, line 133: verifiable

src/transaction.ts
```diff
@@ -234,6 +234,10 @@ export class Transaction implements ISignable {
     serializeForSigning(signedBy: Address): Buffer {
         // TODO: for appropriate tx.version, interpret tx.options accordingly and sign using the content / data hash
         let plain = this.toPlainObject(signedBy);
+        // Make sure we never sign the transaction with another signature set up (useful when using the same method for verification)
+        if (plain.signature) {
```
bogdan-rosianu, line 238: shouldn't this be part of `toPlainObject()` ?

bogdan-rosianu (changes requested):

src/transaction.ts
ccorcoveanu: no, we should be able to convert `toPlainObject` while still keeping the signature. The sig removal should only be done while verifying the signature

danielailie (approved):

bogdan-rosianu (approved):

PR #33, opened 2021-07-20
Login token

arhtudormorar:


src/dapp/walletProvider.ts
```diff
@@ -134,7 +134,9 @@ export class WalletProvider implements IDappProvider {
 
             window.addEventListener('message', connectUrl);
         }).then((connectionUrl: string) => {
-            window.location.href = `${this.baseWalletUrl()}${connectionUrl}?callbackUrl=${options !== undefined && options.callbackUrl !== undefined ? options.callbackUrl : window.location.href}`;
+            const callbackUrl = `callbackUrl=${options !== undefined && options.callbackUrl !== undefined ? options.callbackUrl : window.location.href}`;
+            const token = options !== undefined && options.token !== undefined ? `&token=${options.token}` : '';
+            window.location.href = `${this.baseWalletUrl()}${connectionUrl}?${callbackUrl}${token}`;
```
ccorcoveanu, line 139: I think if we do it something like this it would be even more readable:

```js
let callbackUrl = `callbackUrl=${window.location.href}`;
if (options && options.callbackUrl) {
  callbackUrl = `callbackUrl=${options.callbackUrl}`;
}

let token = '';
if (options && options.token) {
  token = `&token=${options.token}`;
}
```

ccorcoveanu (changes requested):
Could you please edit the `Changelog.md` and add un the [Unreleased] section a reference to this PR and feature?

src/dapp/walletProvider.ts
arhtudormorar: OK

ccorcoveanu (approved):

constantintovisi (approved):

sfeather:
Can you please tell me what does the "token" parameter do?
I can't find it anywhere in the documentation (via https://docs.elrond.com/wallet/webhooks/).

Thank you!

PR #47, opened 2021-08-04
Avoid changing global axios.defaults.transformResponse

pctj101:
What do you think about isolating the axios config to the local file, rather than changing axios for the global scope?

For example, this is one case of the axios global config change introduced by erdjs that conflicts with arweave.
https://github.com/ElrondNetwork/elrond-sdk-erdjs/issues/46

It could make it easier to integrate elrond/erdjs with more dapps with fewer chances of conflict.

### DIFF
--- a/src/apiProvider.ts
+++ b/src/apiProvider.ts
@@ -22,11 +22,17 @@ export class ApiProvider implements IApiProvider {
      * @param url the URL of the Elrond Api
      * @param config axios request config options
      */
-    constructor(url: string, config?: AxiosRequestConfig) {
-        this.url = url;
-        this.config = config || {
-            timeout: 1000,
-        };
+     constructor(url: string, config?: AxiosRequestConfig) {
+      this.url = url;
+      // See: https://github.com/axios/axios/issues/983
+      this.config = config || {
+        timeout: 1000,
+        transformResponse: [
+          function(data) {
+            return JSONbig.parse(data);
+          },
+        ],
+      };
     }
 
     /**
@@ -90,10 +96,3 @@ export class ApiProvider implements IApiProvider {
         throw new errors.ErrApiProviderGet(resourceUrl, originalErrorMessage, error);
     }
 }
-
-// See: https://github.com/axios/axios/issues/983
-axios.defaults.transformResponse = [
-    function (data) {
-        return JSONbig.parse(data);
-    },
-];
--- a/src/apiProvider.ts
+++ b/src/apiProvider.ts
@@ -22,17 +22,17 @@ export class ApiProvider implements IApiProvider {
      * @param url the URL of the Elrond Api
      * @param config axios request config options
      */
-     constructor(url: string, config?: AxiosRequestConfig) {
+    constructor(url: string, config?: AxiosRequestConfig) {
       this.url = url;
-      // See: https://github.com/axios/axios/issues/983
-      this.config = config || {
+      this.config = Object.assign({}, config, {
         timeout: 1000,
+        // See: https://github.com/axios/axios/issues/983 regarding transformResponse
         transformResponse: [
-          function(data) {
+          function(data: any) {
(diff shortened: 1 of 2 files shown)


ccorcoveanu:
Good catch, not the best choice to make global changes to imported libraries. Will review your PR and definitely change this approach

src/apiProvider.ts
```diff
@@ -24,10 +24,16 @@ export class ApiProvider implements IApiProvider {
    */
    constructor(url: string, config?: AxiosRequestConfig) {
     this.url = url;
+    // See: https://github.com/axios/axios/issues/983
```
ccorcoveanu, line 27: Maybe tie the comment to the `transformResponse` property?

src/proxyProvider.ts
```diff
@@ -26,8 +26,14 @@ export class ProxyProvider implements IProvider {
      */
     constructor(url: string, config?: AxiosRequestConfig) {
         this.url = url;
+        // See: https://github.com/axios/axios/issues/983
         this.config = config || {
```
ccorcoveanu, line 30: I guess we should use `Object.assign` or something to merge the provided config and the defaults so that one default won't get lost when a `config` is passed as a parameter.

ccorcoveanu (changes requested):

src/proxyProvider.ts
ccorcoveanu: Ping @pctj101 - Do you think you can implement the requested changes? Or should we do it - there's no problem either way.
Then we can merge this PR :)

pctj101: Oh hi!  I was caught up in a few things but I can get to it later today :)  Thanks for saying hi!

src/apiProvider.ts
```diff
@@ -24,10 +24,16 @@ export class ApiProvider implements IApiProvider {
...
    constructor(url: string, config?: AxiosRequestConfig) {
     this.url = url;
-    this.config = config || {
+    this.config = Object.assign({}, config, {
       timeout: 1000,
-    };
-}
+      // See: https://github.com/axios/axios/issues/983 regarding transformResponse
+      transformResponse: [
+        function(data: any) {
```
pctj101, line 31: Typescript compiling complained about a lack of type for data, so I put "any"

It seems in most cases it's a string, but looking at axios itself, it seems there are times that even axios needs to fall back when data is not a string.  

https://github.com/axios/axios/blob/master/lib/defaults.js#L72

At least for elrond cases, the config scoped for the purposes of elrond APIs seems to be able to assume a string.

What do you think?

src/proxyProvider.ts
pctj101: pushed up a change, but definitely take a look at the data type comment below in case you have any insight.

src/apiProvider.ts
ccorcoveanu: Hey there. It's ok for now to leave it with `any`. Regardless of what `axios` provides in `transformResponse` as `data`, I think we should make sure ourselves it's a string before passing it to `JSONbig.parse`. But this is out of the scope of this PR, I'll make sure we update the error handling here in the case `data` is not a string.

src/proxyProvider.ts
ccorcoveanu: Thanks a lot!

ccorcoveanu (approved):

src/apiProvider.ts
ccorcoveanu: @pctj101 I changed the base to development. Can you please resolve the two conflicts? They're small, you should just keep your changes :)

pctj101: @ccorcoveanu Hello again!  Appreciate the feedback.  Merged against development and matched up the indent.

ccorcoveanu (approved):

radumojic (approved):

PR #48, opened 2021-08-10
called setAddress on ledger login

bogdan-rosianu:
Called `setAddress` when logging with Ledger with the specified index.
Also, removed `addressIndex` from the constructor

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ Breaking changes:
 -   Changed how test wallets (alice, bob, carol etc.) are obtained, added all 12 test wallets:
     - Removed `TestWallets`
     - Use added function `loadTestWallets` (or `loadInteractive`)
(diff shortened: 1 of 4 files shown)


SebastianMarian (dismissed):

ccorcoveanu (changes requested):
Please add an entry in the changelog, unreleased section, with a reference to this PR and note the breaking changes it introduces (you have an example in the release of v6.0.0)

ccorcoveanu:
@bogdan-rosianu The unreleased section is badly formatted. I know it's not your PR that affected this, but can we do it here so we won't end up merging it like this?
I will green this PR anyway in case we decide to fix the changelog in a follow up PR

ccorcoveanu (approved):

ccorcoveanu (approved):

SebastianMarian (approved):

PR #51, opened 2021-08-11
Fixed the address on refresh - main

radumojic:


ioandanpopa (dismissed):

src/dapp/walletConnectProvider.ts
```diff
@@ -43,8 +43,15 @@ export class WalletConnectProvider implements IDappProvider {
           this.walletConnector.connected &&
           this.walletConnector.accounts.length
         ) {
-          const [account] = this.walletConnector.accounts;
-          this.loginAccount(account);
+            const [account] = this.walletConnector.accounts;
+            if (account.includes(".")) {
+                const address = account.split(".");
```
ccorcoveanu, line 48: Can you make this line `const [address, signature] = account.split..`? Because below we pass `address[0]` and `address[1]` and the second param is not an address - becomes confusing

radumojic: updated

constantintovisi (approved):

ccorcoveanu (approved):

PR #59, opened 2021-08-18
Extension provider

andreigiura:
Implements extension provider


ccorcoveanu (changes requested):
The implementation looks nice. Some small comments, but the main reason for the request changes review is:
- Can you please add an entry to the `Changelog` in the `Unreleased` section regarding this PR? You already have examples there, it's pretty straight forward

src/dapp/extensionProvider.ts
```diff
@@ -0,0 +1,212 @@
...
+    signOnly: boolean
+  ): Promise<Array<Transaction>> {
+    this.openExtensionPopup();
+    const data = {
+      from: this.account.index,
+      transactions: transactions,
+      signOnly,
+    };
+
+    return await this.startExtMsgChannel("transaction", data);
```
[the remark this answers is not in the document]
andreigiura: Actually, the same value is used for sign and send transactions, the only difference is the signOnly  flag. So changing it to signTransaction is not descriptive. I am open to suggestions.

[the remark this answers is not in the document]
ccorcoveanu: I would have two endpoints, I am not a big fan of having options of the type (`if flagIsSet-> doSomeOtherStuff`). This opens the door for other flags, and it can get messy. Then we'd need a refactor, which will be a breaking change and so on :)

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how
       - ESDT system smart contract
     - Added `BalanceBuilder` interface
     - Added `NativeSerializer`
+    - Added `ExtensionProvider`
```
ccorcoveanu, line 17: This is actually added in the `Contract Wrapper` PR section.
We should add a new section, under everything from `Fixed ledger login`:

```
- [Fixed Ledger login](https://github.com/ElrondNetwork/elrond-sdk-erdjs/pull/48)
   - Fixed Ledger login by also setting the index to the device.
  
   - ### Breaking changes:
     - Removed `addressIndex` from constructor
     - Use `addressIndex` from `options` object of `login` function


- [Added Extension Provider for dapp integration #59](https://github.com/ElrondNetwork/elrond-sdk-erdjs/pull/59)
```

Or something like that

src/dapp/extensionProvider.ts
```diff
@@ -0,0 +1,214 @@
...
+              this.handleExtResponseErr(event);
+              this.extensionPopupWindow?.close();
+              window.removeEventListener("message", eventHandler);
+              isResolved = true;
+              resolve(event.data.data);
+              break;
+          }
+        }
+      };
+      var windowCloseInterval = setInterval(() => {
```
ccorcoveanu, line 189: Can we please change this to `const`?

src/dapp/extensionProvider.ts
```diff
@@ -0,0 +1,214 @@
...
+      window.addEventListener("message", eventHandler, false);
+    });
+  }
+
+  private handleExtResponseErr(event: any) {
+    if (!event.data || !event.data.data) {
+      if (
+        event.data.type === "transactionComplete" &&
+        event.data.data.length === 0
+      )
```
ccorcoveanu, line 207: Can we use brackets on all `if` conditions? It will be aligned with the rest of the code, also my linter does not like the missing `{}`

src/dapp/extensionProvider.ts
```diff
@@ -0,0 +1,214 @@
...
+            reject("Extension window was closed without response.");
+        }
+      }, 500);
+
+      window.addEventListener("message", eventHandler, false);
+    });
+  }
+
+  private handleExtResponseErr(event: any) {
+    if (!event.data || !event.data.data) {
```
ccorcoveanu, line 203: Is this condition correct?
`if (!event.data) {...we-are-checking event.data.type}`

I also think it will be cleaner with early exits (less conditions and indentation)
```js
if (!havingCorrectCondition) {
    return;
}

if (....) {
    throw;
}
```

ccorcoveanu (approved):

bogdan-rosianu (approved):

PR #60, opened 2021-08-23
Wallet provider no iframe

ccorcoveanu:


src/dapp/walletProvider.ts
```diff
@@ -1,240 +1,97 @@
...
-                }
-
-                clearTimeout(timeout);
-                window.removeEventListener('message', logout.bind(this));
-                return resolve(data.data);
-            };
-
-            window.addEventListener('message', logout);
-        });
+        window.location.href = `${this.baseWalletUrl()}${WALLET_PROVIDER_DISCONNECT_URL}?${callbackUrl}`;
```
bogdan-rosianu, line 86: can you please extract a function that composes these URLs and make unit tests for them?

src/dapp/walletProvider.ts
```diff
@@ -359,6 +164,47 @@ export class WalletProvider implements IDappProvider {
         throw new ErrNotImplemented();
     }
 
+    static isTxSignReturnSuccess(urlParams: any): boolean {
+        return urlParams.hasOwnProperty(WALLET_PROVIDER_CALLBACK_PARAM) && urlParams[WALLET_PROVIDER_CALLBACK_PARAM] === WALLET_PROVIDER_CALLBACK_PARAM_TX_SIGNED;
+    }
+
+    static getTxSignReturnValue(urlParams: any): Transaction[] {
+        const expectedProps = ["nonce", "value", "receiver", "sender", "gasPrice",
+            "gasLimit", "data", "chainID", "version", "signature"];
```
bogdan-rosianu, line 173: `options` ?

bogdan-rosianu (changes requested):

src/dapp/walletProvider.ts
ccorcoveanu: The actual wallet does not take `options` into consideration and never sends it back. This would be another task, starting with the wallet hooks, then adapting `erdjs`

src/dapp/walletProvider.ts
ccorcoveanu: Did not extract into a function since it's a one liner, but managed to unit tests redirect urls (the trick here was to be able to mock `window.location.href`)

bogdan-rosianu (approved):

arhtudormorar (dismissed):

bogdan-rosianu (approved):

arhtudormorar (approved):

PR #62, opened 2021-08-24
Added signTransactions to the WalletConnectProvider

radumojic:


ioandanpopa (approved):

src/dapp/walletConnectProvider.ts
```diff
@@ -170,6 +170,33 @@ export class WalletConnectProvider implements IDappProvider {
...
+     * Signs an array of transactions and returns it
+     * @param transactions
+     */
+    async signTransactions(transactions: Transaction[]): Promise<Transaction[]> {
+        if (!this.walletConnector) {
+            Logger.error("signTransactions: Wallet Connect not initialised, call init() first");
+            throw new Error("Wallet Connect not initialised, call init() first");
+        }
+
+        const address = await this.getAddress();
```
ccorcoveanu, line 183: I really don't like that `getAddress` which is on the interface returns a `string` instead of an `Address` but I guess we'll do this in another PR since it's unrelated :) Just wanted to touch this :)

ccorcoveanu (approved):

PR #74, opened 2021-09-22
Don't allow undefined in Address constructor

lcswillems:
Currently this code is allowed:

```
new Address(undefined);
```

But this code doesn't make sense and is bug-prone. I discovered several bugs in my code because I passed `undefined` to `Address` constructor.


lcswillems:
Hi @andreibancioiu,

You can do as you wish. This is not my lib. If I opened this PR, it is because I believe it is not a good practice to allow undefined. And I don't get in which case `new Address(undefined)` would make sense?

danielailie:
I will close this as this doesn't apply anymore.
Thank you for you report

PR #75, opened 2021-09-23
signing a transaction using hash

bogdan-rosianu:
This PR shows the way to integrate signing transactions over their hash if version=2 and options=1. possible implementation + unit test. 
The resulting transaction and it's signature can be checked in the following test (from elrond-go):
`ElrondNetwork/elrond-go/integrationTests/frontend/wallet/txInterception_test.go.TestInterceptedTxWithSigningOverTxHash`


popenta:
Transaction signing by hash has been done in [this PR](https://github.com/multiversx/mx-sdk-js-core/pull/406).

PR #76, opened 2021-09-27
fix timeout passed by config being ignored

StanislavSava:
This PR fixes the issue where passing a config with a timeout to the ProxyProvider constructor gets ignored with the hardcoded 1000ms timeout.

### DIFF
--- a/src/proxyProvider.ts
+++ b/src/proxyProvider.ts
@@ -16,6 +16,16 @@ const JSONbig = require("json-bigint");
 /**
  * This will be deprecated once all the endpoints move to ApiProvider
  */
+
+const defaultConfig ={
+    timeout: 1000,
+    // See: https://github.com/axios/axios/issues/983 regarding transformResponse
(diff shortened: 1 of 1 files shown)


src/proxyProvider.ts
```diff
@@ -16,6 +16,16 @@ const JSONbig = require("json-bigint");
 /**
  * This will be deprecated once all the endpoints move to ApiProvider
  */
+
+const defaultConfig ={
```
ccorcoveanu, line 20: Just a small typo, space is missing after `=` => `= {`

src/proxyProvider.ts
```diff
@@ -1,4 +1,6 @@
 import axios, { AxiosRequestConfig } from "axios";
+import BigNumber from "bignumber.js";
```
mrazvanalex, line 2: I just don't see this used anywhere. Is this required?

PR #81, opened 2021-10-12
Set offset when reading integers from buffer

kraus-milan:
Older version of Buffer (4.9.2) used in webpack 4 throws error when reading integers without specifying offset.


raresserban:
I also have experience this issue and would like this fix to be included in the sdk.

mihaieremia (approved):

dragos-rebegea (approved):

ccorcoveanu (approved):

PR #92, opened 2021-11-14
Apply user configuration for ProxyProvider last

MWFIAE:
The configuration provided by the user should take precedence over the elrond provided default.
Right now for example the user can't change the timeout to something greater than 1000.


andreibancioiu:
@MWFIAE, very good idea :bulb: 

Sorry for the delay :pray: In the meantime, the constructors of the providers have been changed, and follow the pattern described by you:

```
constructor(url: string, config?: AxiosRequestConfig) {
    this.url = url;
    this.config = { ...defaultConfig, ...config };
}
```

PR #95, opened 2021-12-16
fix browser minified output file

juliancwirko:
There is a problem with minified erdjs for the browser. For the demo from here: 

https://gist.github.com/juliancwirko/f31aab85d735d7b86162de4c4da33b1f

When I use `erdjs.min.js` instead of `erdjs.js`, I will get the error. See attached image.

bug

After using terser instead of tinyify seems to work well, the compression level is a little bit better, and I think terser is a lot faster.

I've tried to run the browser tests. Some of them fail, but the same is on the development branch.


andreibancioiu:
@juliancwirko, thank you for the contribution :tada: 

The browser tests issues should be resolved in this PR: https://github.com/ElrondNetwork/elrond-sdk-erdjs/pull/101.

Furthermore, thank you for the suggestion on using `terser` :rocket: . 

Though, do you mind if we keep your PR on hold a little bit more (until we fix the issues around browserify & latest versions of wallet connect)?

juliancwirko:
Thanks for the response, and sure, do what is necessary :)

PR #98, opened 2022-01-13
Handle heterogeneous enums with fields

andreibancioiu:
(Pair programming, @danielailie and @andreibancioiu)

 - `typesystem/enum` and `codec/enum` (and `typeMapper`) have been adjusted to support (heterogeneous) enums with fields.
 - `typesystem/structs` and `typesystem/enum` have been refactored to rely on `typesystem/fields`.
 - `codec/structs` and `codec/enum` have been refactored to rely on `codec/fields`.
 - `AsyncCall` ABI type is handled as `Nothing` (`NothingType`, `NothingValue`).

**Breaking changes**
 - `ErrStructTyping` has been removed
 - `EnumValue.valueOf()` does not return a string (the name of the variant) anymore. It returns a JS object such as `{"name": "name of variant", "fields": { ... }}`
 - `StructFieldDefinition` has been replaced by `FieldDefinition`, while `StructField` has been replaced by `Field`.

### DIFF
--- a/src/smartcontracts/codec/nothing.ts
+++ b/src/smartcontracts/codec/nothing.ts
@@ -1,0 +1,19 @@
+import { NothingValue } from "../typesystem";
+
+export class NothingCodec {
+    decodeNested(): [NothingValue, number] {
+        return [new NothingValue(), 0];
+    }
+
+    decodeTopLevel(): NothingValue {
+        return new NothingValue();
+    }
+
+    encodeNested(): Buffer {
+        return Buffer.from([]);
+    }
+
+    encodeTopLevel(): Buffer {
+        return Buffer.from([]);
+    }
+}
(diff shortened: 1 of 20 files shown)


src/smartcontracts/codec/nothing.ts
```diff
@@ -0,0 +1,19 @@
+import { NothingValue } from "../typesystem";
+
+export class NothingCodec {
```
ccorcoveanu, line 3: What's the purpose of this class? Why do we need it? ( Or, to put it deeper: What's the purpose of nothing? :) )

andreibancioiu: There are some types (in the ABI and / or in the Rust framework) encoded as nothing (no bytes) / require no decoding - e.g. the type `AsyncCall` (present both in the framework and in the ABIs),  `SCResult<()>` (this one only exists in the Rust framework).

`NothingType`, `NothingValue` have been added to handle these special types of the ABI.

CC: @andrei-marinica

ccorcoveanu (approved):

claudiu725 (approved):

PR #109, opened 2022-01-31
Hotfix: sign transaction with Wallet Connect

phanletrunghieu:
## Issue

Screen Shot 2022-02-01 at 02 07 44

Maiar app sometime return the signature in an array, so the we need to catch this case.

## Reviewers

@arhtudormorar @bogdan-rosianu


ccorcoveanu:
Closed this PR: 
1. The wallet-connect provider has a separate repo now
2. Not actually an issue, as an array of sigs will be returned only on `signTransactions`

PR #114, opened 2022-02-03
Fix recursivity issue when mapping types

claudiu725:
This PR fixes `Error: type isn't known: TokenIdentifier` when trying to decode the output of the `getAllAuctions` endpoint.
I added two basic tests for `TokenIdentifier` in `binary.spec.ts` and in `typeMapper.spec.ts`, as I initially thought the issue was with the `TokenIdentifier` type. However, this was not the case. I think it's fine if the tests stay though.
As such, I created a test in `esdt.spec.ts` which uses the data from the original report, decoding the output of the `getAllAuctions` endpoint from pre-defined data. I had to update `esdt-nft-marketplace.abi.json` to the most recent version and add the `getAllAuctions` endpoint description manually.

**The issue explained**
The issue occurs because the `EsdtToken` type is located below the `Auction` type in the json ABI. I confirmed this by running the tests with those types in reversed order, which allows the tests to pass.
After the ABI is loaded the type remapping occurs in `remapToKnownTypes` from `abiRegistry.ts`, which calls `mapType` from `typeMapper.ts` in order to resolve the actual type of the types. As an example, after loading the the json ABI, the object which represents the token identifier type is `new Type("TokenIdentifier", typeParameters)`. Remapping it results re-creates it as `new TokenIdentifierType("TokenIdentifier", typeParameters)`. As a side note, this was also the reason for the ambiguous `Error: type isn't known: TokenIdentifier`, because the check `instanceof PrimitiveType` from `onTypedValueSelect` fails for `Type`, but not for `TokenIdentifier` which inherits from `PrimitiveType`.
The remapping step made the assumption that all the types resulting from `let knownClosedType = this.closedTypesMap.get(type.getName());` are fully resolved, which is not always true if we have a nested structure where the structs are loaded in reverse order. In this case `Auction` contains `EsdtToken`, which contains `TokenIdentifier`. The latter was left unresolved, since the handling was stopping one nesting level short, hence the error.

Also fixed the guard messages when endpoints are not found.

### DIFF
--- a/src/smartcontracts/typesystem/typeMapper.ts
+++ b/src/smartcontracts/typesystem/typeMapper.ts
@@ -97,7 +97,7 @@ export class TypeMapper {
         }
     }
 
-    mapType(type: Type): Type {
+    mapRecursiveType(type: Type): Type | null {
         let isGeneric = type.isGenericType();
 
         if (type instanceof EnumType) {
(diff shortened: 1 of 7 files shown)


src/smartcontracts/typesystem/typeMapper.ts
```diff
@@ -97,7 +97,7 @@ export class TypeMapper {
         }
     }
 
-    mapType(type: Type): Type {
+    mapRecursiveType(type: Type): Type | null {
```
andreibancioiu, line 100: :+1:

andreibancioiu (approved):

danielailie (approved):
:+1:

PR #120, opened 2022-02-08
Fix type of token.supply

andreibancioiu:
Fix type of `token.supply` (`number` instead of `string`).


claudiulataretu (approved):

andreibancioiu:
Not needed anymore.

PR #123, opened 2022-02-09
Fix transaction.getAsOnNetwork() to work both with ProxyProvider and ApiProvider (PARTIAL FIX)

andreibancioiu:
Should partially fix https://github.com/ElrondNetwork/elrond-sdk-erdjs/issues/96.

However, there are large discrepancies between a `TransactionOnNetwork` returned by the Proxy (Gateway) and a `TransactionOnNetwork` as return from the API - especially related to smart contract results. For example, the API returns the `data` field as a base64-encoded string (leading to parsing failures), while the Proxy (Gateway) returns the decoded string (as expected in the past). This discrepancies should be handled in a future PR (most probably we will merge `ApiProvider` & `ProxyProvider` into a single provider).


src/smartcontracts/interaction.local.net.spec.ts
```diff
@@ -12,12 +12,12 @@ import { Balance } from "../balance";
...
 import { NetworkConfig } from "../networkConfig";
 import { BytesValue } from "./typesystem/bytes";
-import { chooseProvider } from "../interactive";
+import { chooseProxyProvider } from "../interactive";
 
 
 describe("test smart contract interactor", function () {
     let checker = new StrictChecker();
-    let provider = chooseProvider("local-testnet");
+    let provider = chooseProxyProvider("local-testnet");
```
andreigiura, line 20: this should throw an error

danielailie (approved):

PR #130, opened 2022-02-17
Extra exports, better toString(), better handling of empty addresses, gasLimit helpers

andreibancioiu:
- Export `asyncTimer` and `transactionWatcher`. 
 - Export `logger` (so that SDK clients can set the log level).
 - Fix `Address.toJSON()`, `Address.hex()`, `Address.bech32()` when address is empty (less constraints).
 - Override `toString()` for `NumericalValue` and `TokenIdentifier`.
 - Added `Interactor.withGasLimitComponents()` and `GasLimit.add()` helper functions.


src/address.ts
```diff
@@ -182,6 +188,10 @@ export class Address {
      * Converts the address to a pretty, plain JavaScript object.
      */
     toJSON(): object {
+        if (this.isEmpty()) {
```
ccorcoveanu, line 191: Maybe we don't need this? I think it would be better to have the `bech32` and `pubkey` properties - they will be empty with the rest of the changes. Better than users getting `cannot read property of undefined` errors

andreibancioiu: You are right, missed that :rocket:
Fixed :)

ccorcoveanu (approved):

danielailie (approved):

claudiu725 (approved):

PR #140, opened 2022-03-02
Fix padding for `getTxFieldsForEsdtTransfer`

claudiu725:
- The `getTxFieldsForEsdtTransfer` function was hex-encoding the amount, but was not zero-padding the hex string when the resulting string's length was odd. This PR fixes the issue.
- De-duplicated padding code from `getHexMagnitudeOfBigInt`

### DIFF
--- a/src/smartcontracts/codec/utils.ts
+++ b/src/smartcontracts/codec/utils.ts
@@ -1,4 +1,5 @@
 import BigNumber from "bignumber.js";
+import { numberToPaddedHex } from "../../utils.codec";
 
 /**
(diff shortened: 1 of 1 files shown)


src/utils.codec.ts
```diff
@@ -1,5 +1,7 @@
-export function numberToPaddedHex(value: number) {
-    let hex = value.toString(16);
+import BigNumber from "bignumber.js";
+
+export function numberToPaddedHex(value: BigNumber.Value) {
```
andreibancioiu, line 3: :+1:

andreibancioiu (approved):

danielailie (approved):

PR #141, opened 2022-03-03
Added payableBySc to CodeMetadata

MWFIAE:


src/smartcontracts/codeMetadata.ts
```diff
@@ -93,5 +108,6 @@ enum ByteZero {
 
 enum ByteOne {
     Reserved1 = 1,
-    Payable = 2
+    Payable = 2,
+    PayableBySc = 4
```
andreibancioiu, line 112: I can confirm this is correct :heavy_check_mark: 

https://github.com/ElrondNetwork/elrond-vm-common/blob/master/codeMetadata.go#L18

andreibancioiu (approved):
Thank you for your contribution :pray: :tada:

PR #145, opened 2022-03-10
Remove JavaScript's native instanceof, where possible

andreibancioiu:
Related to:
 - https://github.com/ElrondNetwork/elrond-sdk-erdjs/pull/128
 - https://github.com/ElrondNetwork/elrond-sdk-erdjs/pull/143

Usages of `instanceof` lead to fragile type matching and type checking in erdjs' typesystem. In this PR we replace most occurrences of JavaScript's `instanceof` operator with a custom approach, tailored to erdjs's typesystem.

This isn't a breaking change.

### DIFF
--- a/src/smartcontracts/typesystem/types.ts
+++ b/src/smartcontracts/typesystem/types.ts
@@ -21,6 +21,17 @@ export class Type {
         return this.name;
     }
 
+    /**
+     * Gets the fully qualified name of the type, to allow for better (efficient and non-ambiguous) type comparison within erdjs' typesystem.
+     */
+    getFullyQualifiedName(): string {
+        let joinedTypeParameters = this.getTypeParameters().map(type => type.getFullyQualifiedName()).join(", ");
+
+        return this.isGenericType() ? 
+            `erdjs:types:${this.getName()}<${joinedTypeParameters}>` : 
+            `erdjs:types:${this.getName()}`;
+    }
+
     getTypeParameters(): Type[] {
         return this.typeParameters;
     }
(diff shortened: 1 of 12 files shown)


src/smartcontracts/typesystem/types.ts
```diff
@@ -194,7 +247,7 @@ export abstract class PrimitiveValue extends TypedValue {
 }
 
 export function isTyped(value: any) {
-    return value instanceof TypedValue;
+    return value.belongsToTypesystem;
```
claudiu725, line 250:
```suggestion
    return value.belongsToTypesystem !== undefined;
```
`isTyped` should return `true`/`false` rather than a function or `undefined`.

claudiu725 (changes requested):

claudiu725 (approved):

ccorcoveanu (approved):

paul-v-87:
I am currently getting bit by this change in a stack with Ember.js where part of the bundle is generated with webpack, which minifies the code using `terser`. By default `terser` mangles the class names.

Of course, this breaks code like this: `type.hasJavascriptConstructor(VariadicType.name)`, because the `name` property of that class is definitely not the original one anymore. There is quite a large (and quite heated) discussion about this here: https://github.com/webpack-contrib/uglifyjs-webpack-plugin/issues/269

My current fix is to disable mangling in Terser's options (https://github.com/terser/terser#minify-options), but this is something that is quite hard to pin down and should be at least documented somewhere.

PR #152, opened 2022-03-16
Fix data fetching from the blockchain - wait before polling for the first time

ctindogarus4f:
**Current behaviour:**

The polling happens right away, so the user gets the `TransactionWatcher.awaitConditionally(): cannot (yet) fetch data.` error in almost all the cases.

**Desired behaviour:**
Instead, we should do the first polling after X milliseconds so we can guarantee a higher success rate when fetching the data from the blockchain.

See https://github.com/ElrondNetwork/elrond-sdk-erdjs/issues/151 for the full context.

### DIFF
--- a/src/transactionWatcher.ts
+++ b/src/transactionWatcher.ts
@@ -102,6 +102,8 @@ export class TransactionWatcher {
         });
 
         while (!stop) {
+            await periodicTimer.start(this.pollingInterval);
+
             try {
                 fetchedData = await doFetch();
                 Logger.debug("TransactionWatcher.awaitConditionally(): fetched data.", this.hash.toString())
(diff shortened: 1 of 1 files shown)


andreibancioiu (dismissed):
:rocket: 

Nice contribution :pray:

andreibancioiu:
@ctindogarus4f, do you think you can use signed commits? (I am not sure whether you can revert your already pushed commit and sign it).

claudiu725 (dismissed):

ctindogarus4f:
> @ctindogarus4f, do you think you can use signed commits? (I am not sure whether you can revert your already pushed commit and sign it).

Sure, I've just signed the commit now. 👍 

The only way to do this was to force push the previous commit, so the approvals got wiped out from the PR. Feel free to re-check it now :D

andreibancioiu (approved):

claudiulataretu (approved):

ctindogarus4f:
Thanks guys. 

Can you please close the issue associated with this PR when this gets released?  https://github.com/ElrondNetwork/elrond-sdk-erdjs/issues/151

PR #176, opened 2022-04-01
Breaking change: simplify network config / improve design - not a singleton anymore.

andreibancioiu:
- Removed `NetworkConfig.getDefault()` and `NetworkConfig.sync()`. Instead, one should use `let networkConfig = await provider.getNetworkConfig()`.
 - Constructor of `Transaction` now requires `chainID`, as well.
 - Added `Interaction.withChainID()` - must be used before calling `buildTransaction()`.

### DIFF
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -139,7 +139,7 @@ export class Transaction implements ISignable {
     this.value = value || Balance.Zero();
     this.sender = sender || Address.Zero();
     this.receiver = receiver;
-    this.gasPrice = gasPrice || GasPrice.default();
+    this.gasPrice = gasPrice || GasPrice.min();
     this.gasLimit = gasLimit;
     this.data = data || new TransactionPayload();
     this.chainID = chainID;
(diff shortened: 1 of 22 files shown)


src/transaction.ts
```diff
@@ -223,6 +223,11 @@ export class Transaction implements ISignable {
     return this.chainID;
   }
 
+  setChainID(chainID: ChainID) {
+    this.chainID = chainID;
+    this.doAfterPropertySetter();
```
ccorcoveanu, line 228: I think this is a side effect that the user does not expect - I remember debugging with @arhtudormorar quite a lot what was the problem with a transaction, and it was missing the signature after setting some properties. 
If a user changes something on a signed transaction, the signature will be invalid even if it's set or if it's empty, so I don't know if we should force the reset for them

andreibancioiu: Fixed, removed `doAfterPropertySetter()`.

claudiu725 (approved):

ccorcoveanu (approved):

PR #180, opened 2022-04-05
Breaking change: unifying provider interfaces, preparing network providers for extraction - step 1

andreibancioiu:
First part of decoupling `src/networkProvider` from erdjs components. 

When decoupling is done, no file within `src/networkProvider` should reference files from other packages. Then, we can extract `networkProvider` to a separate repository and continue the interface reconciliation process (API & Proxy).


src/networkProvider/transactionStatus.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    /**
+     * Returns whether the transaction is pending (e.g. in mempool).
+     */
+    isPending(): boolean {
+        return (
+            this.status == "received" ||
+            this.status == "pending" ||
+            this.status == "partially-executed"
```
bogdan-rosianu, line 31: can we have status `partially-executed`? I cannot find it here:
https://github.com/ElrondNetwork/elrond-go-core/blob/main/data/transaction/status.go
and I can only find it in a unit test on elrond-proxy-go

bogdan-rosianu (approved):

ccorcoveanu (approved):

PR #184, opened 2022-04-06
Breaking change: unifying provider interfaces, preparing network providers for extraction - step 4

andreibancioiu:
Third part of decoupling `src/networkProvider` from erdjs components: move `queryResponse` to `networkProvider` package.

When decoupling is done, no file within `src/networkProvider` should reference files from other packages. Then, we can extract `networkProvider` to a separate repository and continue the interface reconciliation process (API & Proxy).

### DIFF
--- a/src/networkProvider/constants.ts
+++ b/src/networkProvider/constants.ts
@@ -1,0 +1,3 @@
+import BigNumber from "bignumber.js";
+
+export const MaxUint64 = new BigNumber("18446744073709551615");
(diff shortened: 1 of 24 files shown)


src/networkProvider/constants.ts
```diff
@@ -0,0 +1,3 @@
+import BigNumber from "bignumber.js";
+
+export const MaxUint64 = new BigNumber("18446744073709551615");
```
bogdan-rosianu, line 3: can you rename it to `MaxUint64AsBigNumber` ?

bogdan-rosianu (approved):

bogdan-rosianu (approved):

src/networkProvider/constants.ts
andreibancioiu: Fixed.

danielailie (approved):

PR #190, opened 2022-04-09
Breaking changes: cleanup and minor improvements prior release (step 1)

andreibancioiu:
- erdjs does not depend on `axios` and `fs` anymore. Changed the way an `AbiRegistry` is created. Removed function `Code.fromFile()` (was relying on `fs`).
 - Removed `boundaryAdapters` (simplification).
 - Removed `smartContract.getOwner()` (surprising behavior of function, usage not recommended).
 - Removed opaque, quite redundant verification (`ErrTransactionNotSigned`).
 - Removed `transaction.awaitHashed()` (lacks use-case).
 - Removed `transaction.awaitSigned()` (it isn't useful in relation with any of our signing providers).
 - Removed opaque side-effect (contract address computed upon signing the deploy transaction - not applicable when using our signing providers, anyway).
 - Removed not used interfaces (e.g. `ISignable` is not needed in erdjs, but in walletcore).
 - Removed not used fields: `contract.code`, `contract.codeMetadata`.
 - Removed `interaction.getContract()`. Add `interaction.getContractAddress()`.
 - Removed not-needed interface `IResultsParser`.
 - Removed not needed function: `Code.nothing().`


bogdan-rosianu (approved):

src/smartcontracts/interface.ts
```diff
@@ -16,7 +15,7 @@ export interface ISmartContract {
     /**
      * Gets the address of the Smart Contract.
      */
-    getAddress(): Address;
+    getAddress(): IBech32Address;
```
ccorcoveanu, line 18: Q - why do we have IBech32 as opposed to an interface for the address?

andreibancioiu: It was more like a choice to avoid ambiguity. That is, `IBech32Address` only has the `bech32()` method, without `pubkey()` and `hex()`, being thus being a bit more "portable" between different domains (e.g. `walletcore`, `network-providers` and so on).

We can change this, though. `IAddress`?

andreibancioiu: Renamed to `IAddress`.

ccorcoveanu (approved):

bogdan-rosianu (approved):

ccorcoveanu (approved):

bogdan-rosianu (approved):

ccorcoveanu (approved):

PR #192, opened 2022-04-10
Prepare new beta pre-release

andreibancioiu:
- Decouple (somehow) `Balance` from `Token`.
 - Prepare new beta pre-release.

### DIFF
--- a/src/balance.ts
+++ b/src/balance.ts
@@ -100,6 +100,7 @@ export class Balance {
     }
 
     // TODO: We should not keep a property of a token instance (its nonce) here, in the "Tokens" (still called "Balance") class.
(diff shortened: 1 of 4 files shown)


src/balance.ts
```diff
@@ -89,6 +99,7 @@ export class Balance {
         };
     }
 
+    // TODO: We should not keep a property of a token instance (its nonce) here, in the "Tokens" (still called "Balance") class.
```
bogdan-rosianu, line 102: also, I think the number of decimals has to be passed as well into this component, if we want to use it for different tokens

bogdan-rosianu (approved):

bogdan-rosianu (approved):

claudiu725 (approved):

PR #221, opened 2022-06-28
added a builder for relayed v2 transactions

bogdan-rosianu:
added a builder for relayed v2 txs.
tested on testnet by using the resulted transaction from the unit test:
https://testnet-explorer.elrond.com/transactions/c02398ab9f02e564539ef1d8acd0c36cb4b8720d4c3397ef2e2ed656177471ff

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2622,20 +2622,6 @@
       "integrity": "sha512-OO0pH2lK6a0hZnAdau5ItzHPI6pUlvI7jMVnxUQRtw4owF2wk8lOSabtGDCTP4Ggrg2MbGnWO9X8K1t4+fGMDw==",
       "dev": true
     },
-    "node_modules/fsevents": {
-      "version": "2.3.2",
-      "resolved": "https://registry.npmjs.org/fsevents/-/fsevents-2.3.2.tgz",
-      "integrity": "sha512-xiqMQR4xAeHTuB9uWm+fFRcIOgKBMiOBP+eXiyT7jsgVCq1bkVygt00oASowB7EdtpOHaaPgKt812P9ab+DDKA==",
-      "dev": true,
-      "hasInstallScript": true,
-      "optional": true,
-      "os": [
-        "darwin"
-      ],
-      "engines": {
-        "node": "^8.16.0 || ^10.6.0 || >=11.0.0"
-      }
-    },
     "node_modules/function-bind": {
       "version": "1.1.1",
       "resolved": "https://registry.npmjs.org/function-bind/-/function-bind-1.1.1.tgz",
@@ -8006,13 +7992,6 @@
       "resolved": "https://registry.npmjs.org/fs.realpath/-/fs.realpath-1.0.0.tgz",
       "integrity": "sha512-OO0pH2lK6a0hZnAdau5ItzHPI6pUlvI7jMVnxUQRtw4owF2wk8lOSabtGDCTP4Ggrg2MbGnWO9X8K1t4+fGMDw==",
       "dev": true
-    },
-    "fsevents": {
-      "version": "2.3.2",
-      "resolved": "https://registry.npmjs.org/fsevents/-/fsevents-2.3.2.tgz",
-      "integrity": "sha512-xiqMQR4xAeHTuB9uWm+fFRcIOgKBMiOBP+eXiyT7jsgVCq1bkVygt00oASowB7EdtpOHaaPgKt812P9ab+DDKA==",
-      "dev": true,
-      "optional": true
     },
     "function-bind": {
       "version": "1.1.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,7 @@
         "buffer": "6.0.3",
         "json-duplicate-key-handle": "1.0.0",
         "keccak": "3.0.1",
-        "protobufjs": "6.10.2"
+        "protobufjs": "^6.11.3"
       },
       "devDependencies": {
         "@elrondnetwork/erdjs-network-providers": "0.1.4",
@@ -4396,9 +4396,9 @@
       "dev": true
     },
     "node_modules/protobufjs": {
-      "version": "6.10.2",
-      "resolved": "https://registry.npmjs.org/protobufjs/-/protobufjs-6.10.2.tgz",
-      "integrity": "sha512-27yj+04uF6ya9l+qfpH187aqEzfCF4+Uit0I9ZBQVqK09hk/SQzKa2MUqUpXaVa7LOFRg1TSSr3lVxGOk6c0SQ==",
+      "version": "6.11.3",
+      "resolved": "https://registry.npmjs.org/protobufjs/-/protobufjs-6.11.3.tgz",
+      "integrity": "sha512-xL96WDdCZYdU7Slin569tFX712BxsxslWwAfAhCYjQKGTq7dAU91Lomy6nLLhh/dyGhk/YH4TwTSRxTzhuHyZg==",
       "hasInstallScript": true,
       "dependencies": {
         "@protobufjs/aspromise": "^1.1.2",
@@ -4412,7 +4412,7 @@
         "@protobufjs/pool": "^1.1.0",
         "@protobufjs/utf8": "^1.1.0",
         "@types/long": "^4.0.1",
-        "@types/node": "^13.7.0",
+        "@types/node": ">=13.7.0",
         "long": "^4.0.0"
       },
       "bin": {
@@ -9330,9 +9330,9 @@
       "dev": true
     },
     "protobufjs": {
-      "version": "6.10.2",
-      "resolved": "https://registry.npmjs.org/protobufjs/-/protobufjs-6.10.2.tgz",
-      "integrity": "sha512-27yj+04uF6ya9l+qfpH187aqEzfCF4+Uit0I9ZBQVqK09hk/SQzKa2MUqUpXaVa7LOFRg1TSSr3lVxGOk6c0SQ==",
+      "version": "6.11.3",
+      "resolved": "https://registry.npmjs.org/protobufjs/-/protobufjs-6.11.3.tgz",
+      "integrity": "sha512-xL96WDdCZYdU7Slin569tFX712BxsxslWwAfAhCYjQKGTq7dAU91Lomy6nLLhh/dyGhk/YH4TwTSRxTzhuHyZg==",
       "requires": {
         "@protobufjs/aspromise": "^1.1.2",
         "@protobufjs/base64": "^1.1.2",
@@ -9345,7 +9345,7 @@
         "@protobufjs/pool": "^1.1.0",
         "@protobufjs/utf8": "^1.1.0",
         "@types/long": "^4.0.1",
-        "@types/node": "^13.7.0",
+        "@types/node": ">=13.7.0",
         "long": "^4.0.0"
       }
     },
--- a/package-lock.json
+++ b/package-lock.json
@@ -119,13 +119,13 @@
       "dev": true
     },
     "node_modules/@babel/generator": {
-      "version": "7.18.6",
-      "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.18.6.tgz",
-      "integrity": "sha512-AIwwoOS8axIC5MZbhNHRLKi3D+DMpvDf9XUcu3pIVAfOHFT45f4AoDAltRbHIQomCipkCZxrNkfpOEHhJz/VKw==",
+      "version": "7.18.7",
+      "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.18.7.tgz",
+      "integrity": "sha512-shck+7VLlY72a2w9c3zYWuE1pwOKEiQHV7GTUbSnhyl5eu3i04t30tBY82ZRWrDfo3gkakCFtevExnxbkf2a3A==",
       "dev": true,
       "dependencies": {
-        "@babel/types": "^7.18.6",
-        "@jridgewell/gen-mapping": "^0.3.0",
+        "@babel/types": "^7.18.7",
+        "@jridgewell/gen-mapping": "^0.3.2",
         "jsesc": "^2.5.1"
       },
       "engines": {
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -33,7 +33,7 @@
     "buffer": "6.0.3",
     "json-duplicate-key-handle": "1.0.0",
     "keccak": "3.0.1",
-    "protobufjs": "6.10.2"
+    "protobufjs": "^6.11.3"
```
andreibancioiu, line 36: Perhaps we should drop the caret and keep a fixed version (as for the rest)?

src/relayedTransactionV2Builder.spec.ts
```diff
@@ -0,0 +1,41 @@
+import {loadTestWallets, TestWallet} from "./testutils";
```
andreibancioiu, line 1: The formatting is unlike other imports (from other files).

src/relayedTransactionV2Builder.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+            15,
+            new TransactionPayload("getContractConfig"),
+            60_000_000,
+            "T",
+            networkConfig);
+        await alice.signer.sign(relayedTxV2);
+
+        assert.equal(
+            relayedTxV2.getData().toString(),
+            "relayedTxV2@000000000000000000010000000000000000000000000000000000000002ffff@0f@676574436f6e7472616374436f6e666967@b6c5262d9837853e2201de357c1cc4c9803988a42d7049d26b7785dd0ac2bd4c6a8804b6fd9cf845fe2c2a622774b1a2dbd0a417c9a0bc3f0563a85bd15e710a");
```
andreibancioiu, line 37: :+1:

src/relayedTransactionV2Builder.ts
```diff
@@ -0,0 +1,59 @@
...
+     * @param {IAddress} innerTransactionReceiver The recipient address of the inner transaction
+     * @param {INonce} innerTransactionNonce The nonce of the inner transaction
+     * @param {ITransactionPayload} innerTransactionData The data field of the inner transaction
+     * @param {IGasLimit} innerTransactionGasLimit The gas limit of the inner transaction. It will be used for the big transaction
+     * @param {IChainID} chainID The Chain ID of the inner transaction
+     * @param {INetworkConfig} networkConfig The network config to be used for computing the gas units
+     * @returns {Transaction} Transaction containing the relayed transaction
+     */
+    public async buildRelayedTransactionV2(
+        innerTransactionSenderAccount: ISigner,
```
andreibancioiu, line 23: Would it be possible to allow the caller provide the inner transaction as an argument, instead of these fields as arguments?

src/relayedTransactionV2Builder.ts
```diff
@@ -0,0 +1,59 @@
...
+        networkConfig: INetworkConfig): Promise<Transaction> {
+        let innerTransaction = new Transaction({
+            nonce: innerTransactionNonce,
+            receiver: innerTransactionReceiver,
+            gasLimit: 0,
+            data: innerTransactionData,
+            chainID: chainID,
+        });
+
+        await innerTransactionSenderAccount.sign(innerTransaction);
```
andreibancioiu, line 38: Unfortunately, this will not be appropriate for most dApps using dapp-core (they do not generally use the `ISigner` abstraction). Would it be possible to receive an already-signed transaction in this function (asserting for `gasLimit == 0`)?

src/relayedTransactionV2Builder.ts
```diff
@@ -0,0 +1,59 @@
...
+     * @param {ISigner} innerTransactionSenderAccount The signer of the inner transaction
+     * @param {IAddress} innerTransactionReceiver The recipient address of the inner transaction
+     * @param {INonce} innerTransactionNonce The nonce of the inner transaction
+     * @param {ITransactionPayload} innerTransactionData The data field of the inner transaction
+     * @param {IGasLimit} innerTransactionGasLimit The gas limit of the inner transaction. It will be used for the big transaction
+     * @param {IChainID} chainID The Chain ID of the inner transaction
+     * @param {INetworkConfig} networkConfig The network config to be used for computing the gas units
+     * @returns {Transaction} Transaction containing the relayed transaction
+     */
+    public async buildRelayedTransactionV2(
```
andreibancioiu, line 22: How about:

```
builder
    .setInnerTransaction(signedInnerTransactionWithoutGasLimit as Transaction)
    .setInnerTransactionGasLimit(12345)
    .setNetworkConfig(...)
    .build();
    
```

src/relayedTransactionV2Builder.ts
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * Generate a relay v2 transaction to pay the gas for our users.
+     *
+     * @param {ISigner} innerTransactionSenderAccount The signer of the inner transaction
+     * @param {IAddress} innerTransactionReceiver The recipient address of the inner transaction
+     * @param {INonce} innerTransactionNonce The nonce of the inner transaction
+     * @param {ITransactionPayload} innerTransactionData The data field of the inner transaction
+     * @param {IGasLimit} innerTransactionGasLimit The gas limit of the inner transaction. It will be used for the big transaction
+     * @param {IChainID} chainID The Chain ID of the inner transaction
+     * @param {INetworkConfig} networkConfig The network config to be used for computing the gas units
```
andreibancioiu, line 19: Normally, the network config includes the chain ID, as well.

package-lock.json
```diff
@@ -16,7 +16,7 @@
         "buffer": "6.0.3",
         "json-duplicate-key-handle": "1.0.0",
         "keccak": "3.0.1",
-        "protobufjs": "6.10.2"
```
mariusmihaic, line 19: [comment]

I'm not familiar with js. Did you update this package manually, or does it automatically update?

src/relayedTransactionV2Builder.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+
+describe("test relayed v2 transaction", function () {
+    let alice: TestWallet, bob: TestWallet;
+
+    before(async function () {
+        ({alice, bob} = await loadTestWallets());
+    });
+
+    it("should compute relayed v2 tx", async function () {
+        this.timeout(20000);
```
mariusmihaic, line 15: [question]

Is this 20 seconds? If so, why such long time? Is it an integration test?

src/tokenTransferBuilders.ts
```diff
@@ -47,7 +47,7 @@ export class ESDTNFTTransferPayloadBuilder {
             BytesValue.fromUTF8(this.payment.tokenIdentifier),
             // The nonce of the token
             new U64Value(this.payment.nonce),
-            // The transfered quantity
+            // The transferred quantity
```
mariusmihaic, line 50: :+1:

src/relayedTransactionV2Builder.ts
mariusmihaic: :+1:

src/relayedTransactionV2Builder.ts
mariusmihaic: :+1:

src/relayedTransactionV2Builder.ts
```diff
@@ -0,0 +1,59 @@
...
+                new BytesValue(innerTransactionData.valueOf()),
+                BytesValue.fromHex(innerTransaction.getSignature().hex()),
+            ])
+            .build();
+
+        return new Transaction({
+            receiver: innerTransaction.getSender(),
+            value: 0,
+            gasLimit:
+                innerTransactionGasLimit.valueOf() + networkConfig.MinGasLimit + networkConfig.GasPerDataByte * payload.length(),
```
mariusmihaic, line 54: Could we move this logic in `https://github.com/ElrondNetwork/elrond-sdk-erdjs/blob/main/src/gasEstimator.ts` ?

src/relayedTransactionV2Builder.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+            Address.fromBech32("erd1qqqqqqqqqqqqqqqpqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqzllls8a5w6u"),
+            15,
+            new TransactionPayload("getContractConfig"),
+            60_000_000,
+            "T",
+            networkConfig);
+        await alice.signer.sign(relayedTxV2);
+
+        assert.equal(
+            relayedTxV2.getData().toString(),
```
mariusmihaic, line 36:
```suggestion
        assert.equal(relayedTxV2.getData().toString(),
```

src/relayedTransactionV2Builder.ts
```diff
@@ -0,0 +1,59 @@
+import {Transaction} from "./transaction";
+import {TransactionPayload} from "./transactionPayload";
+import {AddressValue, BytesValue, ContractFunction, U64Value} from "./smartcontracts";
+import {IAddress, IChainID, IGasLimit, INonce, ITransactionPayload} from "./interface";
+import {ISigner} from "@elrondnetwork/erdjs-walletcore/out/interface";
+import {INetworkConfig} from "./interfaceOfNetwork";
+
+export class RelayedTransactionV2Builder {
```
mariusmihaic, line 8: Allt tx builders in this repo are defined here: https://github.com/ElrondNetwork/elrond-sdk-erdjs/blob/main/src/transactionFactory.ts. Could we move the logic here to keep consistency?

src/relayedTransactionV2Builder.ts
mariusmihaic: Right, allow the caller to have both options

src/relayedTransactionV2Builder.spec.ts
bogdan-rosianu: fixed

src/relayedTransactionV2Builder.ts
bogdan-rosianu: sent the inner tx, which has to be already signed and with gas limit not 0

src/relayedTransactionV2Builder.ts
bogdan-rosianu: thanks, refactored

src/relayedTransactionV2Builder.ts
bogdan-rosianu: done :rocket:

src/relayedTransactionV2Builder.ts
bogdan-rosianu: done

package-lock.json
bogdan-rosianu: I have updated it because it was an warning raised by `npm audit` which is automatically run when installing the dependencies

src/relayedTransactionV2Builder.spec.ts
bogdan-rosianu: it is easier to read if there are more lines

src/relayedTransactionV2Builder.ts
bogdan-rosianu: it is rather more similar to `tokenTransferBuilders.ts` so I will leave it as it is

src/relayedTransactionV2Builder.ts
bogdan-rosianu: I wouldn't add an extra dependency to this class. Also, it is pretty specific to this use-case

mariusmihaic: That's exactly what `gasEstimator` does. Computes gas for every specific use-case. 
I would keep the consistency, we could also ask @andreibancioiu 's opinion on this

src/relayedTransactionV2Builder.ts
mariusmihaic: `tokenTransferBuilders` builds the payLoad. `transactionFactory` builds the txs. You could've split `relayedTransactionV2Builder` in 3 to keep consistency:
- gasEstimator for relayed tx
- payloadBuilder for relayed tx
- txFactory to create the relayed tx.

Again, let's also hear @andreibancioiu 's opinion

src/relayedTransactionV2Builder.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+        };
+
+        const innerTx = new Transaction({
+            nonce: 15,
+            receiver: Address.fromBech32("erd1qqqqqqqqqqqqqqqpqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqzllls8a5w6u"),
+            gasLimit: 10,
+            chainID: networkConfig.ChainID,
+            data: new TransactionPayload("getContractConfig"),
+        });
+        builder = builder.setNetworkConfig(networkConfig).setInnerTransactionGasLimit(10).setInnerTransaction(innerTx);
```
andreibancioiu, line 38: :+1:

src/relayedTransactionV2Builder.ts
andreibancioiu: The need for _consistency_ is indeed a very good argument. On the other hand, this new component is self-contained, which is also good.

If we were to move `gasEstimator.ts`, `tokenTransferBuilders.ts` and `transactionFactory.ts` to a new package called `transfers` (in the future), then this design is a good choice. It's possible that in the future we might even extract the following logic outside of erdjs (and move it to separate packages):
 - transfers (EGLD and all kinds of ESDT)
 - relayed transactions (V1 & V2)
 - staking & delegation SC calls (actually, missing here now)
 - esdt management calls (actually, missing here now)

So, no strong opinion at this moment (can stay as it is if we desire to further split erdjs functionality to separate NPM packages),

src/relayedTransactionV2Builder.ts
```diff
@@ -0,0 +1,81 @@
...
+     *
+     * @throws ErrInvalidRelayedV2BuilderArguments
+     * @throws ErrGasLimitShouldBe0ForInnerTransaction
+     * @return Transaction
+     */
+    build() : Transaction {
+        if(!this.innerTransaction || !this.innerTransactionGasLimit || !this.netConfig || !this.innerTransaction.getSignature()) {
+            throw new ErrInvalidRelayedV2BuilderArguments();
+        }
+        if(this.innerTransaction.getGasLimit() !== 0){
```
andreibancioiu, line 58: Comparison might fail if `getGasLimit` returns an object that implements `IGasLimit`. Perhaps allow type coercion to happen? `!=` / or add a test with both numbers and objects if this works as it is.

src/relayedTransactionV2Builder.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+            MinGasLimit: 50_000,
+            GasPerDataByte: 1_500,
+            GasPriceModifier: 0.01,
+            ChainID: "T"
+        };
+
+        const innerTx = new Transaction({
+            nonce: 15,
+            receiver: Address.fromBech32("erd1qqqqqqqqqqqqqqqpqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqzllls8a5w6u"),
+            gasLimit: 10,
```
andreibancioiu, line 34: Using some implementation of `IGasLimit`, such as `{ valueOf: function() { return 10; } }` might fail the test.

src/relayedTransactionV2Builder.ts
andreibancioiu: (added a separate comment)

Both options are valid and both are backed by good arguments :balance_scale:

src/relayedTransactionV2Builder.spec.ts
bogdan-rosianu: extended the test to include this as well

mariusmihaic (approved):

PR #223, opened 2022-07-12
fix: Update type of abiRegistry types

maximechbt:
The type of the type parameters is not correct, it must be an object and not an array.
this fix is ​​not perfect, it would be better to define a more precise type, but that would already unlock the use.


ccorcoveanu (changes requested):
Hello, thanks dor the update! Could you please point the PR to `feat/erdjs-v12`?

ccorcoveanu (approved):

PR #228, opened 2022-07-14
Fix type expression parser

nvergez:
Hello, I got an error when I tried to import and parse an abi with the type `tuple<List<u64>,List<u64>>`. I saw the commented test in `src/smartcontracts/typesystem/typeExpressionParser.spec.ts` which is basically the same case as mine.

I just fixed the problem adding a condition in the function `getJsonedString`. I uncommented the commented test and added my specific case in the test file because it tests the '`>` before `,`' case and the 'JSON key de-duplication' case.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -65,30 +65,30 @@
       }
     },
     "node_modules/@babel/compat-data": {
-      "version": "7.18.6",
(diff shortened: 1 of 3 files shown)


ccorcoveanu (changes requested):
@nvergez can you please run an npm install and also commit package-lock?

ccorcoveanu (approved):

andreibancioiu (approved):

PR #239, opened 2022-09-29
Guarded transaction

schimih:


src/networkParams.ts
```diff
@@ -34,8 +34,8 @@ export class TransactionVersion {
     /**
      * Creates a TransactionVersion object with the VERSION setting for hash signing
      */
-    static withTxHashSignVersion(): TransactionVersion {
-        return new TransactionVersion(TRANSACTION_VERSION_TX_HASH_SIGN);
+    static withTxOptions(): TransactionVersion {
```
ccorcoveanu, line 37: Like above, we should document this change

src/transaction.ts
```diff
@@ -200,13 +226,22 @@ export class Transaction {
...
     // Make sure we never sign the transaction with another signature set up (useful when using the same method for verification)
     if (plain.signature) {
       delete plain.signature;
     }
+
+    if (plain.guardianSignature) {
+      delete plain.guardianSignature;
+    }
+
+    if ((plain.guardian == undefined) || (plain.guardian == "")) {
```
ccorcoveanu, line 241: We should use `===`, because... Javascript. But I think any falsy value should delete the property, so you could use `if (!plain.guardian)`

src/transaction.ts
```diff
@@ -149,6 +163,10 @@ export class Transaction {
     return this.receiver;
   }
 
+  getGuardian(): IAddress | undefined {
```
ccorcoveanu, line 166: Can this be undefined though? Given this line in the constructor: `this.guardian = guardian || Address.empty();`

ccorcoveanu: Also, seeing the logic in the serializer that throws when transaction options say `it's a guarded tx`, but the `guardianSignature === 0x000[..]00`, we should either:
1. Throw here in the constructor when not providing a guardian (we don't have a setter for it, so if it's undefined, the transaction is invalid forever)
2. Add a setter and let the serializer go with `Address.empty()`
3. Add a setter and a public methods that checks that the tx has all the guardian fields set up (`return this.getOptions().isGuardedTx() && !this.guardian.equals(Address.empty())`). Then use this method in the serializer check

src/proto/serializer.ts
```diff
@@ -36,15 +36,30 @@ export class ProtoSerializer {
...
+            protoTransaction.GuardSignature = Buffer.from(transaction.getGuardianSignature().hex(), "hex");
+        }
+
+        const encoded = proto.Transaction.encode(protoTransaction).finish();
+        const buffer = Buffer.from(encoded);
+
         return buffer;
     }
 
+    private tryGetGuardianAddress(guardianAddress: IAddress | undefined): IAddress {
```
ccorcoveanu, line 55: Given the comment from `transaction.ts`, I don't think `getGuardian` can return `undefined`. This method could be unnecessary, see comment below

ccorcoveanu (changes requested):

src/constants.ts
```diff
@@ -1,10 +1,14 @@
 export const TRANSACTION_MIN_GAS_PRICE = 1000000000;
 export const TRANSACTION_OPTIONS_DEFAULT = 0;
 export const TRANSACTION_OPTIONS_TX_HASH_SIGN = 1;
+export const TRANSACTION_OPTIONS_TX_GUARDED = 2;
 export const TRANSACTION_VERSION_DEFAULT = 1;
-export const TRANSACTION_VERSION_TX_HASH_SIGN = 2;
```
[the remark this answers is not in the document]
andreibancioiu: Perhaps rename to `TRANSACTION_VERSION_WITH_TX_OPTIONS` (optional).

src/constants.ts
```diff
@@ -1,10 +1,14 @@
...
+export const TRANSACTION_VERSION_TX_OPTIONS = 2;
 export const ESDT_TRANSFER_GAS_LIMIT = 500000;
 export const ESDT_TRANSFER_FUNCTION_NAME = "ESDTTransfer";
 export const ESDTNFT_TRANSFER_FUNCTION_NAME = "ESDTNFTTransfer";
 export const MULTI_ESDTNFT_TRANSFER_FUNCTION_NAME = "MultiESDTNFTTransfer";
 export const ESDT_TRANSFER_VALUE = "0";
+
+// Masks
+export const TRANSACTION_OPTIONS_TX_GUARDED_MASK = 0b0010
```
andreibancioiu, line 14: Shouldn't we have here an additional mask used for `options for tx signed with hash`? CC: @bogdan-rosianu

src/networkParams.ts
andreibancioiu: Documentation can be added here:

https://docs.elrond.com/sdk-and-tools/erdjs/erdjs-migration-guides/

src/networkParams.ts
```diff
@@ -76,6 +76,23 @@ export class TransactionOptions {
         return new TransactionOptions(TRANSACTION_OPTIONS_TX_HASH_SIGN);
     }
 
+    /**
+     * Created a TransactionsOptions object with the setting for guarded transaction
```
andreibancioiu, line 80: Typo: `create` or `creates`.

src/networkParams.ts
```diff
@@ -76,6 +76,23 @@ export class TransactionOptions {
...
+     */
+    static withTxGuardedOptions(): TransactionOptions {
+        return new TransactionOptions(TRANSACTION_OPTIONS_TX_GUARDED);
+    }
+
+    /**
+     * Returns true if Guarded Transaction Option is set
+     */
+    isGuardedTx(): boolean {
```
andreibancioiu, line 89: Do we have a unit test for this? (sorry if we already do).

src/networkParams.ts
```diff
@@ -76,6 +76,23 @@ export class TransactionOptions {
...
+        if ((this.value & TRANSACTION_OPTIONS_TX_GUARDED_MASK) == TRANSACTION_OPTIONS_TX_GUARDED) {
```
andreibancioiu, line 90: Perhaps we can use a shorthand like `return (this.value & TRANSACTION_OPTIONS_TX_GUARDED_MASK) == TRANSACTION_OPTIONS_TX_GUARDED)`.

src/relayedTransactionV2Builder.spec.ts
```diff
@@ -84,7 +84,7 @@ describe("test relayed v2 transaction builder", function () {
         assert.equal(relayedTxV2.getNonce().valueOf(), 37);
         assert.equal(
             relayedTxV2.getData().toString(),
-            "relayedTxV2@000000000000000000010000000000000000000000000000000000000002ffff@0f@676574436f6e7472616374436f6e666967@b6c5262d9837853e2201de357c1cc4c9803988a42d7049d26b7785dd0ac2bd4c6a8804b6fd9cf845fe2c2a622774b1a2dbd0a417c9a0bc3f0563a85bd15e710a");
+            "relayedTxV2@000000000000000000010000000000000000000000000000000000000002ffff@0f@676574436f6e7472616374436f6e666967@bcf4d4aa206c649855aced3168db9684c77bcc0a9a75793bf64fef462bea17afe51625acd11bb9dd84b068fec7661e4d63c2b787f7235f37c237775c9867c105");
```
andreibancioiu, line 87: How come that this changed?

src/smartcontracts/interaction.spec.ts
```diff
@@ -137,7 +137,7 @@ describe("test smart contract interactor", function() {
         assert.equal(transaction.getData().toString(), "getUltimateAnswer");
         assert.equal(
             transaction.getHash().toString(),
-            "60d0956a8902c1179dce92d91bd9670e31b9a9cd07c1d620edb7754a315b4818"
+            "580056f13a82d1df9fed421be7ec3d1024c18fd8f4261c823c0fb3a8766d8561"
```
andreibancioiu, line 140: This should remain unchanged (and the ones below).

src/smartcontracts/interaction.ts
```diff
@@ -182,6 +185,11 @@ export class Interaction {
         return this;
     }
 
+    withSender(sender: IAddress): Interaction {
```
andreibancioiu, line 188: Is this necessary? It does not seem to be used in any test.

schimih: This is used in `system-tests-js` repo.

src/proto/serializer.ts
schimih: true,
removed it
:+1:

src/transaction.ts
schimih: changed

src/constants.ts
bogdan-rosianu: Yes, I think it's ok if we add this

schimih: added 
`export const TRANSACTION_OPTIONS_TX_HASH_SIGN_MASK = 0b0001`

src/relayedTransactionV2Builder.spec.ts
schimih: fixed.
sender was missing

src/networkParams.ts
```diff
@@ -86,15 +86,11 @@ export class TransactionOptions {
     /**
      * Returns true if Guarded Transaction Option is set
      */
-    isGuardedTx(): boolean {
-        if ((this.value & TRANSACTION_OPTIONS_TX_GUARDED_MASK) == TRANSACTION_OPTIONS_TX_GUARDED) {
-            return true
-        }
-        return false
+    withGuardedOptions(): boolean {
```
bogdan-rosianu, line 89: I would rename this function to `hasGuardedOptions`, since `withGuardedOptions` sounds like a "setter"

schimih: changed to `hasGuardedOption`

src/transaction.ts
```diff
@@ -247,11 +249,18 @@ export class Transaction {
     return Buffer.from(serialized);
   }
 
+  /**
+   * Checks the integrity of the guarded transaction if
+   */
+  isGuardedTransaction(): boolean {
+    let guardianComp = this.guardian.bech32() === (Address.empty().bech32());
```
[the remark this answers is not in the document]
schimih: true. changed

bogdan-rosianu (approved):

src/transaction.ts
```diff
@@ -197,40 +227,56 @@ export class Transaction {
...
     // Make sure we never sign the transaction with another signature set up (useful when using the same method for verification)
     if (plain.signature) {
       delete plain.signature;
     }
+
+    if (plain.guardianSignature) {
+      delete plain.guardianSignature;
+    }
+
+    if (!plain.guardian) {
```
andreibancioiu, line 243: I think this attribute should be kept.

src/transaction.ts
```diff
@@ -197,40 +227,56 @@ export class Transaction {
...
     let serialized = JSON.stringify(plain);
 
     return Buffer.from(serialized);
   }
 
+  /**
+   * Checks the integrity of the guarded transaction
+   */
+  isGuardedTransaction(): boolean {
+    const guardianComp = this.guardian.bech32() === (Address.empty().bech32());
```
andreibancioiu, line 256: Rename to `isGuardianSet`?

Below, rename to `isGuardianSignature` set?

src/transaction.ts
```diff
@@ -197,40 +227,56 @@ export class Transaction {
...
       receiver: this.receiver.bech32(),
-      sender: sender ? sender.bech32() : this.sender.bech32(),
+      sender: this.sender.bech32(),
       gasPrice: this.gasPrice.valueOf(),
       gasLimit: this.gasLimit.valueOf(),
       data: this.data.length() == 0 ? undefined : this.data.encoded(),
       chainID: this.chainID.valueOf(),
       version: this.version.valueOf(),
       options: this.options.valueOf() == 0 ? undefined : this.options.valueOf(),
+      guardian: this.guardian?.bech32() ? (this.guardian.bech32() == "" ? undefined : this.guardian.bech32()) : undefined,
```
andreibancioiu, line 277: Very tricky.

andreibancioiu (approved):

ccorcoveanu (approved):

PR #255, opened 2023-02-12
Fixes #254

### LINKED ISSUE #254: Handling Transaction value >= 1e21
The [`toPlainObject`](https://github.com/multiversx/mx-sdk-js-core/blob/e13ad27b48994d1fe72e64b1bce4b81f07e23ba3/src/transaction.ts#L221) and [`fromPlainObject`](https://github.com/multiversx/mx-sdk-js-core/blob/e13ad27b48994d1fe72e64b1bce4b81f07e23ba3/src/transaction.ts#L242) methods of the [`Transaction`](https://github.com/multiversx/mx-sdk-js-core/blob/e13ad27b48994d1fe72e64b1bce4b81f07e23ba3/src/transaction.ts#L23) class in some way calls the `toString()` method of bignumber.js which returns the exponential value for numbers >= 1e21.

This causes invalid signature issues when signing transactions.

newtmex:


andreibancioiu:
Thank you for the contribution :tada: 

Normally, you should not pass `BigNumber` instances to the `Transaction` constructor, but use objects whose `toString()` is suitable on that purpose. E.g., use `TokenPayment` objects:

https://github.com/multiversx/mx-sdk-js-core/blob/main/src/tokenPayment.ts#L65

Currently, we only accept PRs will all commits being signed:

https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

We will discuss a bit this internally, and come with a solution that does not alter the tests. In the meantime, please pass `TokenPayment` objects or `new BigNumber(...).toFixed(0)` as transaction `value` etc. Do not directly pass `BigNumber` objects. That is, do not pass arbitrary objects (`BigNumber` is an arbitrary object in this context).

newtmex:
Thanks for the reply and recommendation.

I will work with your suggestion and wait for the solution the team will come up with.

andreibancioiu:
@newtmex, we've applied the fix here:

https://github.com/multiversx/mx-sdk-js-core/releases/tag/v11.3.0

Thanks a lot :pray: 

For future contributions, make sure to configure commits signing on GitHub.

PR #259, opened 2023-02-18
Temporary, non-breaking workaround: WARN if sender isn't set.

andreibancioiu:
In v12, `sender` will be required on all components.


andreibancioiu:
Actually, we should not apply these changes here, but wait for v12 instead.

andreibancioiu:
Done in v12.

PR #260, opened 2023-02-18
Add a new parameter to the TransactionWatcher: "patience"

andreibancioiu:
_Patience_ = an extra time (in milliseconds) to wait, after the transaction has reached its desired status. 

Reason: currently there's a delay between the moment a transaction is marked as "completed" and the moment its outcome (contract results, events and logs) is available.

**Theoretically, this is a breaking change** - since we've altered the constructor of `TransactionWatcher` (which now receives the options as an object). In practice, most probably it's not a breaking change (since the default configuration may have been rarely altered).


schimih (approved):

andreigiura (approved):

src/transactionWatcher.ts
```diff
@@ -152,6 +161,11 @@ export class TransactionWatcher {
             }
         }
 
+        // The patience timer isn't subject to the timeout constraints.
+        if (satisfied) {
+            await patienceTimer.start(this.patienceMilliseconds);
```
raress96, line 166: Just providing some feedback here, shouldn't `doFetch` be called after waiting for `patienceTimer` so `fetchedData` is actually the up to date data (contains smart contract results/logs etc if that was the usecase) ?

PR #268, opened 2023-03-10
Sketch next SC transactions factory

andreibancioiu:


andreibancioiu:
Closing, since implementation of the new transaction factories happens separately.

PR #275, opened 2023-04-03
Adjust construction of transaction options (breaking change)

andreibancioiu:
- Allow one to create `TransactionOptions` with more than one option, using `TransactionOptions.withOptions()`.
 - Allow one to mutate the options, using `setWithHashSign()`, `setWithGuardian()`

Breaking change (from v11):
 - Remove static function `withTxHashSignOptions`.


src/networkParams.ts
```diff
@@ -63,31 +59,57 @@ export class TransactionOptions {
...
-    static withDefaultOptions(): TransactionOptions {
+    static withDefaultOptions() {
         return new TransactionOptions(TRANSACTION_OPTIONS_DEFAULT);
     }
 
     /**
-     * Creates a TransactionsOptions object with the setting for hash signing
+     * Creates a TransactionOptions object from a set of options.
+     */
+    public static withOptions(options: {
```
andreibancioiu, line 71: Or should we name it `fromFlags()`, `withFlags()`?

src/networkParams.ts
```diff
@@ -63,31 +59,57 @@ export class TransactionOptions {
...
+     * Returns true if the "hash sign" option is set.
+     */
+    isWithHashSign(): boolean {
+        return (this.value & TRANSACTION_OPTIONS_TX_HASH_SIGN) == TRANSACTION_OPTIONS_TX_HASH_SIGN;
+    }
+
+    /**
+     * Returns true if the "guarded transaction" option is set.
      */
-    static withTxHashSignOptions(): TransactionOptions {
```
andreibancioiu, line 75: Or should we leave it, but as deprecated?

src/networkParams.ts
schimih: My opinion is that we should stay with "withOptions( )" given the  `TransactionOptions` name.
Does it make sense to keep the function `withOptions( )` and remove the "with" prefix from the options names?

src/networkParams.ts
schimih: I would remove it.

schimih (commented):
some opinions

schimih (approved):

src/networkParams.ts
andreibancioiu: Indeed. Fixed.

src/networkParams.ts
andreibancioiu: Removed.

bogdan-rosianu (approved):

schimih (approved):

PR #297, opened 2023-06-09
Toggle global burn, registerAndSetAllRoles etc.

andreibancioiu:
- In `TokenOperationsFactory.ts`, add new functions: `registerAndSetAllRoles`, `unsetBurnRoleGlobally`, `setBurnRoleGlobally`. 
 - Functions that create token-registering transactions will emit a WARN to notify the developer about the existence of `unsetBurnRoleGlobally` - to be called, if desired (as a separate transaction). Developers can disable this warning by passing `shouldWarnAboutUnsettingBurnRoleGlobally = false` in the factory's constructor.

### DIFF
--- a/src/tokenOperations/tokenOperationsFactory.ts
+++ b/src/tokenOperations/tokenOperationsFactory.ts
@@ -1,6 +1,7 @@
 import BigNumber from "bignumber.js";
 import { ARGUMENTS_SEPARATOR, TRANSACTION_OPTIONS_DEFAULT, TRANSACTION_VERSION_DEFAULT } from "../constants";
 import { IAddress, IChainID, IGasLimit, IGasPrice, INonce, ITransactionValue } from "../interface";
+import { Logger } from "../logger";
 import { TransactionOptions, TransactionVersion } from "../networkParams";
 import { Transaction } from "../transaction";
 import { TransactionPayload } from "../transactionPayload";
@@ -26,8 +27,6 @@ interface IConfig {
     gasLimitStorePerByte: IGasLimit;
     issueCost: BigNumber.Value;
     esdtContractAddress: IAddress;
-
-    shouldWarnAboutUnsettingBurnRoleGlobally: boolean;
 }
 
 interface IBaseArgs {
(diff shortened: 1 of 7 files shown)


src/tokenOperations/tokenOperationsFactory.ts
```diff
@@ -25,6 +26,8 @@ interface IConfig {
     gasLimitStorePerByte: IGasLimit;
     issueCost: BigNumber.Value;
     esdtContractAddress: IAddress;
+
+    shouldWarnAboutUnsettingBurnRoleGlobally: boolean;
```
andreibancioiu, line 30: In theory, this is a breaking change. However, in practice, it should not be. Generally speaking, users would be using the default config implementation, `TokenOperationsFactoryConfig`. Furthermore, this transactions factory isn't widely used yet.

src/tokenOperations/tokenOperationsFactory.ts
```diff
@@ -12,6 +12,7 @@ interface IConfig {
     minGasLimit: IGasLimit;
     gasLimitPerByte: IGasLimit;
     gasLimitIssue: IGasLimit;
+    gasLimitToggleBurnRoleGlobally: IGasLimit;
```
andreibancioiu, line 15: In theory, this is a breaking change. However, in practice, it should not be. Generally speaking, users would be using the default config implementation, `TokenOperationsFactoryConfig`. Furthermore, this transactions factory isn't widely used yet.

src/tokenOperations/tokenOperationsFactory.ts
```diff
@@ -43,11 +46,19 @@ interface IIssueFungibleArgs extends IBaseArgs {
...
+
+    /**
+     * @deprecated (not used anymore)
+     */
+    canMint?: boolean;
+
+    /**
+     * @deprecated (not used anymore)
+     */
+    canBurn?: boolean;
```
andreibancioiu, line 61: Will be removed in the next major release. Not used anymore.

CiprianDraghici (approved):

src/tokenOperations/tokenOperationsFactory.ts
ccorcoveanu, line 30: I wouldn't use a special setting for a single print. Maybe we can have something more general. I don't really expect we would have more warnings, but still it would look better in my opinion not to have it this specific

andreibancioiu: Fixed, now using our existing `Logger` component, with `INFO` level.

ccorcoveanu (approved):

Issue #283, opened 2023-04-07
Handle "enums" in the type inference system (aka "NativeSerializer")


andreibancioiu:
As a follow-up to: https://github.com/multiversx/mx-sdk-js-core/issues/274

andreibancioiu:
Related: https://github.com/multiversx/mx-sdk-js-core/issues/216

andreibancioiu:
Fixed in https://github.com/multiversx/mx-sdk-js-core/releases/tag/v12.6.0.

Issue #216, opened 2022-05-24
NativeSerializer cannot parse enum value


gfusee:
Hi,

It seems convertToTypedValue method from NativeSerializer cannot handle enum, the workaround is to not currently use ABI to call endpoints that require enum

I'm using erdjs 10.2.4

Thank you

hanahdev:
> Hi,
> 
> It seems convertToTypedValue method from NativeSerializer cannot handle enum, the workaround is to not currently use ABI to call endpoints that require enum
> 
> I'm using erdjs 10.2.4
> 
> Thank you

we are so sorry for the late reply kindly contact there live chat support on whatsapp -  +19546679281

TheoSabattie:
Workaround:
```ts 
axios.get('your.abi.json').then((response) => {
    const networkProvider = new ProxyNetworkProvider(
        'https://devnet-gateway.elrond.com' // or the main/test net
    );

    const abiRegistry = AbiRegistry.create(response.data);
    const abi = new SmartContractAbi(abiRegistry);
    const contract = new SmartContract({
        address: new Address(yourConfig.contractAddress),
        abi: abi
    });

    // here
    const gameStateType = abiRegistry.getEnum('GameState');

    const interaction = contract.methodsExplicit.stateToGames([
      EnumValue.fromName(gameStateType, GameState.WaitingPlayer)
    ]);

    const query = interaction.buildQuery();
    networkProvider.queryContract(query).then((queryResponse) => {
        const endpointDefinition = interaction.getEndpoint();
        const resultsParser = new ResultsParser();

        const { firstValue, secondValue, returnCode } =
            resultsParser.parseQueryResponse(queryResponse, endpointDefinition);
    });
});
```

andreibancioiu:
Fixed in https://github.com/multiversx/mx-sdk-js-core/releases/tag/v12.6.0.

PR #300, opened 2023-06-19
add enums type inference in NativeSerializer and fix Enum valueof()

nvergez:
I have implemented type inference for enum types in the NativeSerializer.

- If a number is provided, it will be processed as the discriminant.
- If a string is provided, it will be processed as the name.
- For enums with fields, a JavaScript object needs to be provided in the following format:
`{ name: string; fields: any}`
 The key of each property in the `fields` object must correspond to the name in the `FieldDefinition`.

I have also updated the `valueof()` method of `EnumValue`. When we receive an ABI file from a smart contract built with complex enums, the names of the fields are the same as those in the Rust code. This means they can be either numbers or strings.

This indicates that the following code cannot work correctly with a string as a name:
https://github.com/multiversx/mx-sdk-js-core/blob/30f4fbfcfaf70b5b30213b5ebb3b9cf5feb6dc1a/src/smartcontracts/typesystem/enum.ts#L150-L152

I believe the fields property should be an object rather than an array, but this change would definitely lead to a breaking change. However, I think my fix doesn't break anything by using the index instead of the name itself.

### DIFF
--- a/src/smartcontracts/typesystem/enum.ts
+++ b/src/smartcontracts/typesystem/enum.ts
@@ -149,7 +149,7 @@ export class EnumValue extends TypedValue {
     valueOf() {
         let result: any = { name: this.name, fields: [] };
 
-        this.fields.forEach((field) => (result.fields[field.name] = field.value.valueOf()));
+        this.fields.forEach((field, index) => (result.fields[index] = field.value.valueOf()));
 
         return result;
     }
--- a/src/smartcontracts/nativeSerializer.spec.ts
+++ b/src/smartcontracts/nativeSerializer.spec.ts
@@ -2,7 +2,7 @@ import BigNumber from "bignumber.js";
 import { assert } from "chai";
 import { Address } from "../address";
 import { NativeSerializer } from "./nativeSerializer";
-import { AbiRegistry, AddressType, AddressValue, BigUIntType, BooleanType, BooleanValue, CompositeType, CompositeValue, EndpointDefinition, EndpointModifiers, EndpointParameterDefinition, ListType, NullType, OptionalType, OptionalValue, OptionType, OptionValue, TupleType, U32Type, U64Type, U64Value, U8Type, U8Value, VariadicType, VariadicValue } from "./typesystem";
+import { AbiRegistry, AddressType, AddressValue, BigUIntType, BooleanType, BooleanValue, CompositeType, CompositeValue, EndpointDefinition, EndpointModifiers, EndpointParameterDefinition, EnumType, ListType, NullType, OptionalType, OptionalValue, OptionType, OptionValue, TupleType, U32Type, U64Type, U64Value, U8Type, U8Value, VariadicType, VariadicValue } from "./typesystem";
 import { BytesType, BytesValue } from "./typesystem/bytes";
 
 describe("test native serializer", () => {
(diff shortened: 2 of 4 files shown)


andreibancioiu (commented):
@nvergez, thanks a lot for the contribution :tada: 

Code looks great :star2: 

This seems to change the behavior of `enum.valueOf()`. Indeed, the previous behavior wasn't very correct (as you have said, it could have been an object, instead). I'll analyze this a bit more, then come back with a re-review.

In the meantime, do you think you can solve the Git conflicts?

src/smartcontracts/typesystem/enum.ts
```diff
@@ -149,7 +149,7 @@ export class EnumValue extends TypedValue {
     valueOf() {
         let result: any = { name: this.name, fields: [] };
 
-        this.fields.forEach((field) => (result.fields[field.name] = field.value.valueOf()));
+        this.fields.forEach((field, index) => (result.fields[index] = field.value.valueOf()));
```
andreibancioiu, line 152: I'll have to analyze this a bit more, then come back.

andreibancioiu:
@nvergez, do you think you can also add the following unit test in `enum.spec.ts`?

The first two assertions should normally hold with your changes, as well. Then, the third one would have to be adjusted, since your code uses indexes to populate `result.fields`.

Indeed, it seems that your code did not change the existing (incorrect) behavior of `enum.valueOf()`, and only extended it to work with properly-named enum fields, as well.

```
it.only("should get valueOf()", () => {
    // Define variants
    let greenVariant = new EnumVariantDefinition("Green", 0, [
        new FieldDefinition("0", "red component", new U8Type()),
        new FieldDefinition("1", "green component", new U8Type()),
        new FieldDefinition("2", "blue component", new U8Type()),
    ]);

    let orangeVariant = new EnumVariantDefinition("Orange", 1, [
        new FieldDefinition("0", "hex code", new StringType())
    ]);

    let yellowVariant = new EnumVariantDefinition("Yellow", 2, [
        new FieldDefinition("red", "red component", new U8Type()),
        new FieldDefinition("green", "green component", new U8Type()),
        new FieldDefinition("blue", "blue component", new U8Type()),
    ]);

    // Define enum type
    let enumType = new EnumType("Colour", [
        greenVariant,
        orangeVariant,
        yellowVariant
    ]);

    // Create enum values
    let green = new EnumValue(enumType, greenVariant, [
        new Field(new U8Value(0), "0"),
        new Field(new U8Value(255), "1"),
        new Field(new U8Value(0), "2")
    ]);

    let orange = new EnumValue(enumType, orangeVariant, [
        new Field(new StringValue("#FFA500"), "0")
    ]);

    let yellow = new EnumValue(enumType, yellowVariant, [
        new Field(new U8Value(255), "red"),
        new Field(new U8Value(255), "green"),
        new Field(new U8Value(0), "blue")
    ]);

    // Test valueOf()
    assert.deepEqual(green.valueOf(), {
        "name": "Green",
        "fields": [
            new BigNumber(0),
            new BigNumber(255),
            new BigNumber(0)
        ]
    });

    assert.deepEqual(orange.valueOf(), {
        "name": "Orange",
        "fields": [
            "#FFA500"
        ]
    });

    // Incorrect (proper field names aren't handled correctly).
    // Normally, fixed by your PR.
    assert.deepEqual(yellow.valueOf(), {
        "name": "Yellow",
        "fields": [
        ]
    });
});
```

In v13, we should definitely have `fields` as an object, thank you :pray:

nvergez:
> In the meantime, do you think you can solve the Git conflicts?

Resolving the conflict will cause a test to fail in this PR because it will add the test from `main` to `nativeSerializer.spec.ts` but not the related fix in `nativeSerializer.ts`.

I'm unsure whether you would prefer to merge main into this branch, or just resolve the conflict and deal with a failing test.

andreibancioiu (approved):
Looks great! :tada:

andreibancioiu:
Should solve:
 - https://github.com/multiversx/mx-sdk-js-core/issues/283
 - https://github.com/multiversx/mx-sdk-js-core/issues/216

Thank you :pray:

schimih (approved):

andreibancioiu:
@nvergez, released here:

https://github.com/multiversx/mx-sdk-js-core/releases/tag/v12.6.0

Thanks a lot for the contribution!

PR #315, opened 2023-08-10
Check if isCompleted is undefined

popenta:
This is done because in the new version of `sdk-network-providers` the `isCompleted` field has become **optional**. Nothing has changed for the `ApiNetworkProvider`, but for the `ProxyNetworkProvider` the `isCompleted` field will only be set if `getTransaction()` is called using `withProcessStatus=true`.

As you may know, the `TransactionWatcher` class needs a `TransactionFetcher` object when initialized. You can use both `ApiNetworkProvider` or `ProxyNetworkProvider` because both respect the interface.

In case you are using the `proxy` provider and you need the `isCompleted` field, a plain object that uses the proxy provider can be passed, defining a `getTransaction()` method that explicitly calls `proxyProvider.getTransaction()` using `withProcessStatus=true`. This can be done as bellow:

```
let provider = new ProxyNetworkProvider("http://localhost:7950", { timeout: 5000 });
let watcher = new TransactionWatcher({
    getTransaction: async (hash: string) => { return await provider.getTransaction(hash, true) }
});
```

Now, the completion checks are more strict, so it takes more time to check if a transaction is completed. The changes have been done in the following PR's:
- https://github.com/multiversx/mx-chain-proxy-go/pull/386
- https://github.com/multiversx/mx-sdk-js-network-providers/pull/41

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
   },
   "devDependencies": {
     "@multiversx/sdk-network-providers": "1.2.1",
+    "@multiversx/sdk-network-providers-next": "npm:@multiversx/sdk-network-providers@1.6.0-beta.0",
     "@multiversx/sdk-wallet": "3.0.0",
     "@multiversx/sdk-wallet-next": "npm:@multiversx/sdk-wallet@4.0.0",
     "@types/assert": "1.4.6",
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -325,3 +325,9 @@ export class ErrGasLimitShouldBe0ForInnerTransaction extends Err {
     super("gas limit must be 0 for the inner transaction for relayed v2");
   }
 }
+
+export class ErrIsCompletedFieldIsMissingOnTransaction extends Err {
+  public constructor() {
+    super("The transaction watcher requires the isCompleted property to be defined on the transaction object. Perhaps you've used the sdk-network-provider's getTransaction() and in that case you should also pass `withProcessStatus=true`.")
+  }
+}
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -325,12 +325,3 @@ export class ErrGasLimitShouldBe0ForInnerTransaction extends Err {
     super("gas limit must be 0 for the inner transaction for relayed v2");
   }
 }
-
-/**
- * Signals an optional class member is undefined
- */
-export class ErrOptionalClassMemberIsUndefined extends Err {
-  public constructor(message: string) {
-    super(message);
-  }
-}
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -325,3 +325,12 @@ export class ErrGasLimitShouldBe0ForInnerTransaction extends Err {
     super("gas limit must be 0 for the inner transaction for relayed v2");
   }
 }
+
+/**
+ * Signals an optional class member is undefined
+ */
+export class ErrOptionalClassMemberIsUndefined extends Err {
+  public constructor(message: string) {
+    super(message);
+  }
+}
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -328,6 +328,6 @@ export class ErrGasLimitShouldBe0ForInnerTransaction extends Err {
 
 export class ErrIsCompletedFieldIsMissingOnTransaction extends Err {
   public constructor() {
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -36,6 +36,7 @@
   },
   "devDependencies": {
     "@multiversx/sdk-network-providers": "1.2.1",
+    "@multiversx/sdk-network-providers-next": "npm:@multiversx/sdk-network-providers@1.6.0-beta.0",
```
andreibancioiu, line 39: All right for now (to reference the beta version), since it's used for tests only (dev dependency). Once `sdk-core` is compatible with both the old and the new `sdk-network-providers` (as of this PR), we can make a proper release there, as well. And afterwards, come back here to adjust the reference.

For `sdk-core`, let's bump de minor version.

src/errors.ts
```diff
@@ -325,3 +325,9 @@ export class ErrGasLimitShouldBe0ForInnerTransaction extends Err {
     super("gas limit must be 0 for the inner transaction for relayed v2");
   }
 }
+
+export class ErrIsCompletedFieldIsMissingOnTransaction extends Err {
+  public constructor() {
+    super("The transaction watcher requires the isCompleted property to be defined on the transaction object. Perhaps you've used the sdk-network-provider's getTransaction() and in that case you should also pass `withProcessStatus=true`.")
```
andreibancioiu, line 331: Perhaps add quotes for `isCompleted`  and `getTransaction()`, as did for `withProcessStatus`.

src/transaction.local.net.spec.ts
```diff
@@ -15,8 +16,8 @@ describe("test transaction", function () {
         ({ alice, bob } = await loadTestWallets());
     });
 
-    it("should send transactions", async function () {
-        this.timeout(30000);
+    it("should send transactions and wait for completion", async function () {
+        this.timeout(70000);
```
andreibancioiu, line 20: Yup. Because the completion checks are now much more strict, https://github.com/multiversx/mx-chain-proxy-go/pull/386, and the wait time is increased.

Let's add links towards the following PRs in the description:
 - https://github.com/multiversx/mx-chain-proxy-go/pull/386
 - https://github.com/multiversx/mx-sdk-js-network-providers/pull/41

src/transaction.local.net.spec.ts
```diff
@@ -62,6 +63,40 @@ describe("test transaction", function () {
         assert.deepEqual(TokenTransfer.egldFromAmount(85).valueOf(), newBalanceOfBob.minus(initialBalanceOfBob));
     });
 
+    it("should send transaction and wait for completion using the new proxy provider", async function () {
+        this.timeout(70000);
+
+        let provider = createLocalnetProvider();
+        let newProvider = new ProxyNetworkProvider("http://localhost:7950", { timeout: 5000 });
+        let watcher = new TransactionWatcher({
+            getTransaction: async (hash: string) => { return await newProvider.getTransaction(hash, true) }
```
andreibancioiu, line 72: This snippet should be added in the PR description (and in the release notes).

src/transactionWatcher.ts
```diff
@@ -155,6 +161,10 @@ export class TransactionWatcher {
             } catch (error) {
                 Logger.debug("TransactionWatcher.awaitConditionally(): cannot (yet) fetch data.");
 
+                if (error instanceof ErrIsCompletedFieldIsMissingOnTransaction) {
```
andreibancioiu, line 164: This is correct.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#conditional_catch_blocks

@CiprianDraghici can confirm, as well.

package.json
popenta: Ok, got it. So after this PR is merged we make a proper release for sdk-network-providers, then come back here and reference the proper release. Will also make a new release for sdk-core increasing the minor version. 👍🏻

bogdan-rosianu (approved):

src/transaction.local.net.spec.ts
popenta: Added in the PR description. Will also add in the release nodes when it's time.

bogdan-rosianu (approved):

andreibancioiu (approved):

PR #328, opened 2023-09-05
Delegation transaction intents factory

popenta:
Implemented the delegation transaction intents factory based on the [sdk-specs](https://github.com/multiversx/mx-sdk-specs/blob/main/sdk-core/transactions-factories/delegation_transaction_intents_factory.md).

### DIFF
--- a/src/transactionIntentsFactories/delegationTransactionIntentsFactory.spec.ts
+++ b/src/transactionIntentsFactories/delegationTransactionIntentsFactory.spec.ts
@@ -36,10 +36,7 @@ describe("test delegation intents factory", function () {
         assert.equal(intent.sender, "erd18s6a06ktr2v6fgxv4ffhauxvptssnaqlds45qgsrucemlwc8rawq553rt2");
         assert.equal(intent.receiver, DELEGATION_MANAGER_SC_ADDRESS);
         assert.isDefined(intent.data);
-
-        let decoder = new TextDecoder();
-        assert.equal(decoder.decode(intent.data), "createNewDelegationContract@010f0cf064dd59200000@0a");
-
+        assert.deepEqual(intent.data, Buffer.from("createNewDelegationContract@010f0cf064dd59200000@0a"));
         assert.equal(intent.gasLimit.valueOf(), 60126500);
         assert.equal(intent.value, value);
     });
@@ -62,10 +59,7 @@ describe("test delegation intents factory", function () {
         assert.equal(intent.sender, "erd18s6a06ktr2v6fgxv4ffhauxvptssnaqlds45qgsrucemlwc8rawq553rt2");
         assert.equal(intent.receiver, "erd1qqqqqqqqqqqqqqqpqqqqqqqqqqqqqqqqqqqqqqqqqqqqqtllllls002zgc");
         assert.isDefined(intent.data);
-
-        let decoder = new TextDecoder();
-        assert.equal(decoder.decode(intent.data), "addNodes@e7beaa95b3877f47348df4dd1cb578a4f7cabf7a20bfeefe5cdd263878ff132b765e04fef6f40c93512b666c47ed7719b8902f6c922c04247989b7137e837cc81a62e54712471c97a2ddab75aa9c2f58f813ed4c0fa722bde0ab718bff382208@81109fa1c8d3dc7b6c2d6e65206cc0bc1a83c9b2d1eb91a601d66ad32def430827d5eb52917bd2b0d04ce195738db216");
-
+        assert.deepEqual(intent.data, Buffer.from("addNodes@e7beaa95b3877f47348df4dd1cb578a4f7cabf7a20bfeefe5cdd263878ff132b765e04fef6f40c93512b666c47ed7719b8902f6c922c04247989b7137e837cc81a62e54712471c97a2ddab75aa9c2f58f813ed4c0fa722bde0ab718bff382208@81109fa1c8d3dc7b6c2d6e65206cc0bc1a83c9b2d1eb91a601d66ad32def430827d5eb52917bd2b0d04ce195738db216"));
         assert.equal(intent.value, 0);
     });
 
@@ -88,10 +82,7 @@ describe("test delegation intents factory", function () {
         assert.equal(intent.sender, "erd18s6a06ktr2v6fgxv4ffhauxvptssnaqlds45qgsrucemlwc8rawq553rt2");
         assert.equal(intent.receiver, "erd1qqqqqqqqqqqqqqqpqqqqqqqqqqqqqqqqqqqqqqqqqqqqqtllllls002zgc");
         assert.isDefined(intent.data);
-
-        let decoder = new TextDecoder();
-        assert.equal(decoder.decode(intent.data), "removeNodes@6e6f746176616c6964626c736b6579686578656e636f646564");
-
+        assert.deepEqual(intent.data, Buffer.from("removeNodes@6e6f746176616c6964626c736b6579686578656e636f646564"));
         assert.equal(intent.value, 0);
     });
 
@@ -114,10 +105,7 @@ describe("test delegation intents factory", function () {
         assert.equal(intent.sender, "erd18s6a06ktr2v6fgxv4ffhauxvptssnaqlds45qgsrucemlwc8rawq553rt2");
         assert.equal(intent.receiver, "erd1qqqqqqqqqqqqqqqpqqqqqqqqqqqqqqqqqqqqqqqqqqqqqtllllls002zgc");
         assert.isDefined(intent.data);
-
-        let decoder = new TextDecoder();
-        assert.equal(decoder.decode(intent.data), "stakeNodes@6e6f746176616c6964626c736b6579686578656e636f646564");
-
+        assert.deepEqual(intent.data, Buffer.from("stakeNodes@6e6f746176616c6964626c736b6579686578656e636f646564"));
         assert.equal(intent.value, 0);
     });
 
@@ -131,19 +119,16 @@ describe("test delegation intents factory", function () {
             }
         };
 
-        const intent = delegationFactory.createTransactionIntentForUnbondingNodes(
-            sender,
-            delegationContract,
-            [publicKey]
-        );
+        const intent = delegationFactory.createTransactionIntentForUnbondingNodes({
+            sender: sender,
+            delegationContract: delegationContract,
+            publicKeys: [publicKey]
+        });
 
         assert.equal(intent.sender, "erd18s6a06ktr2v6fgxv4ffhauxvptssnaqlds45qgsrucemlwc8rawq553rt2");
         assert.equal(intent.receiver, "erd1qqqqqqqqqqqqqqqpqqqqqqqqqqqqqqqqqqqqqqqqqqqqqtllllls002zgc");
         assert.isDefined(intent.data);
-
-        let decoder = new TextDecoder();
-        assert.equal(decoder.decode(intent.data), "unBondNodes@6e6f746176616c6964626c736b6579686578656e636f646564");
-
+        assert.deepEqual(intent.data, Buffer.from("unBondNodes@6e6f746176616c6964626c736b6579686578656e636f646564"));
         assert.equal(intent.value, 0);
         assert.equal(intent.gasLimit.valueOf(), 12143000);
     });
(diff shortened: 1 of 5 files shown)


src/transactionIntentsFactories/delegationTransactionIntentsFactory.ts
```diff
@@ -0,0 +1,344 @@
...
+        return new TransactionIntentBuilder({
+            config: this.config,
+            sender: options.sender,
+            receiver: options.delegationContract,
+            dataParts: dataParts,
+            executionGasLimit: this.computeExecutionGasLimitForNodesManagement(numNodes)
+        }).build();
+    }
+
+    private computeExecutionGasLimitForNodesManagement(numNodes: number): BigNumber.Value {
```
danielailie, line 85: can you please move this at the end of the file or maybe decide on a pattern, first public then private?

danielailie (commented):
because to much red

src/transactionIntentsFactories/delegationTransactionIntentsFactory.spec.ts
```diff
@@ -0,0 +1,301 @@
+import BigNumber from "bignumber.js";
+import { Address } from "../address";
+import { DelegationTransactionIntentsFactory } from "./delegationTransactionIntentsFactory";
+import { assert } from "chai";
+import { DELEGATION_MANAGER_SC_ADDRESS } from "../constants";
+import { ValidatorPublicKey } from "@multiversx/sdk-wallet-next";
+import { SignableMessage } from "../signableMessage";
+
+describe("test delegation intents factory", function () {
+    const delegationFactory = new DelegationTransactionIntentsFactory({
```
andreibancioiu, line 10: If we add a config like here:

https://github.com/multiversx/mx-sdk-py-core/blob/main/multiversx_sdk_core/transaction_intents_factories/transaction_intents_factory_config.py

Then we can also use it in the unit tests (easier).

src/transactionIntentsFactories/delegationTransactionIntentsFactory.spec.ts
```diff
@@ -0,0 +1,301 @@
...
+        gasLimitStake: 5000000,
+        gasLimitUnstake: 5000000,
+        gasLimitUnbond: 5000000,
+        gasLimitCreateDelegationContract: 50000000,
+        gasLimitDelegationOperations: 1000000,
+        additionalGasLimitPerValidatorNode: 6000000,
+        additionalGasLimitForDelegationOperations: 10000000
+    });
+
+    it("should build intent for new delegation contract", async function () {
```
andreibancioiu, line 23: Now I see, we can change "should build" to "should create" (optional, "philosophical").

src/transactionIntentsFactories/delegationTransactionIntentsFactory.spec.ts
```diff
@@ -0,0 +1,301 @@
...
+        const sender = Address.fromBech32("erd18s6a06ktr2v6fgxv4ffhauxvptssnaqlds45qgsrucemlwc8rawq553rt2");
+        const delagationCap = new BigNumber("5000000000000000000000");
+        const serviceFee = new BigNumber(10);
```
andreibancioiu, line 26: We can also use raw strings and numbers, and the factory methods should be able to handle them appropriately. Perhaps do this in some tests? E.g. directly pass `"5000000000000000000000"`.

src/transactionIntentsFactories/delegationTransactionIntentsFactory.spec.ts
```diff
@@ -0,0 +1,301 @@
...
+        const publicKey = new ValidatorPublicKey(Buffer.from("e7beaa95b3877f47348df4dd1cb578a4f7cabf7a20bfeefe5cdd263878ff132b765e04fef6f40c93512b666c47ed7719b8902f6c922c04247989b7137e837cc81a62e54712471c97a2ddab75aa9c2f58f813ed4c0fa722bde0ab718bff382208", "hex"));
+
+        const signedMessage = new SignableMessage();
+        signedMessage.applySignature(Buffer.from("81109fa1c8d3dc7b6c2d6e65206cc0bc1a83c9b2d1eb91a601d66ad32def430827d5eb52917bd2b0d04ce195738db216", "hex"));
+
+        const intent = delegationFactory.createTransactionIntentForAddingNodes({
+            sender: sender,
+            delegationContract: delegationContract,
+            publicKeys: [publicKey],
+            signedMessages: [signedMessage.getSignature()]
```
andreibancioiu, line 56: Since it's a `mock message`, we can do:

```
const mockMessage = {
    getSignature: () => Buffer.from("...", hex)
}
```

Should work. Then, we'll remove `new SignableMessage()` and `applySignature()` from the test.

src/transactionIntentsFactories/delegationTransactionIntentsFactory.spec.ts
```diff
@@ -0,0 +1,301 @@
...
+        assert.deepEqual(intent.data, Buffer.from("addNodes@e7beaa95b3877f47348df4dd1cb578a4f7cabf7a20bfeefe5cdd263878ff132b765e04fef6f40c93512b666c47ed7719b8902f6c922c04247989b7137e837cc81a62e54712471c97a2ddab75aa9c2f58f813ed4c0fa722bde0ab718bff382208@81109fa1c8d3dc7b6c2d6e65206cc0bc1a83c9b2d1eb91a601d66ad32def430827d5eb52917bd2b0d04ce195738db216"));
+        assert.equal(intent.value, 0);
+    });
+
+    it("should build intent for removing nodes", async function () {
+        const sender = Address.fromBech32("erd18s6a06ktr2v6fgxv4ffhauxvptssnaqlds45qgsrucemlwc8rawq553rt2");
+        const delegationContract = Address.fromBech32("erd1qqqqqqqqqqqqqqqpqqqqqqqqqqqqqqqqqqqqqqqqqqqqqtllllls002zgc");
+
+        const publicKey = {
+            hex(): string {
```
andreibancioiu, line 71: :rocket:

src/transactionIntentsFactories/delegationTransactionIntentsFactory.spec.ts
```diff
@@ -0,0 +1,301 @@
...
+                return Buffer.from("notavalidblskeyhexencoded").toString("hex");
```
andreibancioiu, line 72: This seems to be a happy unit test. Perhaps use a valid BLS key, so that there is no ambiguity for the reader?

andreibancioiu: And below, as well? Actually, a dummy value should do, e.g. `abba`, but "notavalidblskeyhexencoded" is somehow misleading.

src/transactionIntentsFactories/delegationTransactionIntentsFactory.ts
```diff
@@ -0,0 +1,344 @@
...
+    gasLimitStake: BigNumber.Value;
+    gasLimitUnstake: BigNumber.Value;
+    gasLimitUnbond: BigNumber.Value;
+    gasLimitCreateDelegationContract: BigNumber.Value;
+    gasLimitDelegationOperations: BigNumber.Value;
+    additionalGasLimitPerValidatorNode: BigNumber.Value;
+    additionalGasLimitForDelegationOperations: BigNumber.Value;
+}
+
+interface ValidatorPublicKey {
```
andreibancioiu, line 23: This is an internal interface, but even so, by convention, we can prefix it with `I`. Indeed, this is not the general practice (e.g. in other projects), but in `sdk-js-*` we did so.

src/transactionIntentsFactories/delegationTransactionIntentsFactory.ts
```diff
@@ -0,0 +1,344 @@
...
+        }
+
+        const numNodes = options.publicKeys.length;
+
+        let dataParts = ["addNodes"];
+        for (let i = 0; i < numNodes; i++) {
+            dataParts = dataParts.concat(options.publicKeys[i].hex());
+            dataParts = dataParts.concat(byteArrayToHex(options.signedMessages[i]));
+        }
+
```
andreibancioiu, line 75: If possible, we can have `const dataParts`, and use [.push()](https://stackoverflow.com/a/35167699).

src/transactionIntentsFactories/delegationTransactionIntentsFactory.ts
```diff
@@ -0,0 +1,344 @@
...
+    createTransactionIntentForRemovingNodes(options: {
+        sender: IAddress,
+        delegationContract: IAddress,
+        publicKeys: ValidatorPublicKey[]
+    }): TransactionIntent {
+        let dataParts = ["removeNodes"];
+
+        for (const key of options.publicKeys) {
+            dataParts = dataParts.concat(key.hex());
```
andreibancioiu, line 98: I am not sure why `concat` works fine in this case. Does the compiler complain if you add the typing markers `dataParts: string[] = ...`? Perhaps use `append`, as mentioned in the comment above?

src/transactionIntentsFactories/delegationTransactionIntentsFactory.ts
```diff
@@ -0,0 +1,344 @@
...
+        publicKeys: ValidatorPublicKey[]
+    }): TransactionIntent {
+        let dataParts = ["unBondNodes"];
+
+        for (const key of options.publicKeys) {
+            dataParts = dataParts.concat(key.hex());
+        }
+
+        const numNodes = options.publicKeys.length;
+        const executionGasLimit = new BigNumber(numNodes).multipliedBy(
```
andreibancioiu, line 150: Can be split, a temporary variable can be used.

src/transactionIntentsFactories/delegationTransactionIntentsFactory.ts
```diff
@@ -0,0 +1,344 @@
...
+
+    createTransactionIntentForUnjailingNodes(options: {
+        sender: IAddress,
+        delegationContract: IAddress,
+        publicKeys: ValidatorPublicKey[]
+    }): TransactionIntent {
+        let dataParts = ["unJailNodes"];
+
+        for (const key of options.publicKeys) {
+            dataParts = dataParts.concat(key.hex());
```
andreibancioiu, line 198: `append` vs. `concat`.

src/transactionIntentsFactories/delegationTransactionIntentsFactory.ts
```diff
@@ -0,0 +1,344 @@
...
+    private readonly config: Config;
+
+    constructor(config: Config) {
+        this.config = config;
+    }
+
+    createTransactionIntentForNewDelegationContract(options: {
+        sender: IAddress,
+        totalDelegationCap: BigNumber.Value,
+        serviceFee: BigNumber.Value,
```
andreibancioiu, line 37: Indeed, we receive as the `BigNumber.Value` interface.

src/transactionIntentsFactories/delegationTransactionIntentsFactory.ts
popenta: Moved to the end of file.

src/transactionIntentsFactories/delegationTransactionIntentsFactory.spec.ts
popenta: Created the config file.

src/transactionIntentsFactories/delegationTransactionIntentsFactory.spec.ts
popenta: Changed to `should create`

src/transactionIntentsFactories/delegationTransactionIntentsFactory.spec.ts
popenta: Changed this test, works as expected.

src/transactionIntentsFactories/delegationTransactionIntentsFactory.spec.ts
popenta: Done

src/transactionIntentsFactories/delegationTransactionIntentsFactory.spec.ts
popenta: Now using `abba` in test

src/transactionIntentsFactories/delegationTransactionIntentsFactory.ts
popenta: Renamed to `IValidatorPublicKey`

src/transactionIntentsFactories/delegationTransactionIntentsFactory.ts
popenta: Now using `push()`. It does not complain if I add typing markers but it already shows the correct type.

src/transactionIntentsFactories/delegationTransactionIntentsFactory.ts
popenta: Done

src/transactionIntentsFactories/delegationTransactionIntentsFactory.ts
popenta: now using `push()`

andreibancioiu (approved):

danielailie (approved):

PR #332, opened 2023-09-17
For constructor of "TokenTransfer", mark "numDecimals" as optional

andreibancioiu:
This allows one to create a `TokenTransfer` without having to explicitly set `numDecimals` to zero if token metadata is not known beforehand or if it simply does not matter in the context (not front-end logic etc.).

The developer can choose to directly call the `TokenTransfer` constructor instead of using the specific named constructors `TokenTransfer.fungibleFromBigInteger`, `TokenTransfer.metaEsdtFromBigInteger` etc. - if token metadata and token type are not known beforehand, and fetching the metadata is not desired in the context.

`TokenTransfer`'s constructor is now public (was protected). Thanks @michavie for reporting this :pray: 

This is not a breaking change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@multiversx/sdk-core",
-  "version": "12.8.0",
+  "version": "12.8.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@multiversx/sdk-core",
-      "version": "12.8.0",
+      "version": "12.8.1",
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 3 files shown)


andreibancioiu:
> I noticed that the `TokenTransfer` class has a `protected` constructor, which restricts direct instantiation. In the past, I've found use cases where I wanted to directly instantiate `TokenTransfer` but had to use `metaEsdtFromBigInteger()` as a workaround.

Good point :raised_hands: Will fix in this PR, thank you :pray:

popenta (dismissed):

popenta (approved):

PR #341, opened 2023-10-24
Transfer transactions factory

popenta:
Implemented the transfer transactions factory as described [here](https://github.com/multiversx/mx-sdk-specs/blob/main/sdk-core/transactions-factories/transfer_transactions_factory.md).

### DIFF
--- a/src/transactionsFactories/transferTransactionsFactory.ts
+++ b/src/transactionsFactories/transferTransactionsFactory.ts
@@ -4,6 +4,7 @@ import { IAddress } from "../interface";
 import { DraftTransaction } from "../draftTransaction";
 import { DraftTransactionBuilder } from "./draftTransactionBuilder";
 import { TokenComputer, NextTokenTransfer } from "../tokens";
+import { ErrBadUsage } from "../errors";
 
 const ADDITIONAL_GAS_FOR_ESDT_TRANSFER = 100000;
 const ADDITIONAL_GAS_FOR_ESDT_NFT_TRANSFER = 800000;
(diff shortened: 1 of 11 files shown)


src/transactionsFactories/transferTransactionsFactory.ts
```diff
@@ -0,0 +1,94 @@
...
+        const numberOfTransfers = options.tokenTransfers.length;
+        let receiver = options.receiver;
+
+        let transferArgs: string[] = [];
+        let extraGasForTransfer = new BigNumber(0);
+        const tokenComputer = new TokenComputer();
+
+        if (numberOfTransfers == 0) {
+            throw new ErrBadUsage("No token transfer has been provided");
+        } else if (numberOfTransfers == 1) {
```
danielailie, line 62: I don't think you need this else here. Also, you need to add === for condition

popenta: indeed, the `else` is not needed. also added `===`.

danielailie (approved):

danielailie (approved):

src/transactionsFactories/transferTransactionsFactory.ts
```diff
@@ -0,0 +1,120 @@
...
+            config: this.config,
+            sender: options.sender,
+            receiver: options.sender,
+            dataParts: transferArgs,
+            gasLimit: extraGasForTransfer,
+            addDataMovementGas: true,
+        }).build();
+    }
+
+    private createSingleESDTTransferDraft(options: {
```
andreibancioiu, line 88: The suffix `draft` can be omitted, just like for the public methods.

andreibancioiu (approved):

PR #349, opened 2023-11-08
Fix applySignature(). Handle buffers in a better way

andreibancioiu:
- Apply the same logic for both messages and transactions.
 - Move `interpretSignatureAsBuffer` to `signature.ts`.

### DIFF
--- a/src/signature.ts
+++ b/src/signature.ts
@@ -1,4 +1,4 @@
-import  * as errors from "./errors";
+import * as errors from "./errors";
 
 
 const SIGNATURE_LENGTH = 64;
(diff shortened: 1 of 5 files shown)


MiroMargineanu (approved):

danielailie (approved):

src/signature.ts
```diff
@@ -58,3 +58,13 @@ export class Signature {
         return this.valueHex;
     }
 }
+
+export function interpretSignatureAsBuffer(signature: { hex(): string; } | Uint8Array): Buffer {
```
MiroMargineanu, line 62: Is the returning type necessary? Seems inferred.

MiroMargineanu (changes requested):

src/signature.ts
andreibancioiu: By convention, on `sdk-core` we add return types (usually).

MiroMargineanu (approved):

PR #353, opened 2023-11-16
Code metadata enhancements

michavie:
- Made properties public since there was no way to read them without parsing from `toJson`
- Added `fromBuffer`
- Added tests

### DIFF
--- a/src/smartcontracts/codeMetadata.ts
+++ b/src/smartcontracts/codeMetadata.ts
@@ -119,13 +119,13 @@ export class CodeMetadata {
     }
 }
 
-enum ByteZero {
+export enum ByteZero {
     Upgradeable = 1,
     Reserved2 = 2,
     Readable = 4
 }
 
-enum ByteOne {
+export enum ByteOne {
     Reserved1 = 1,
     Payable = 2,
     PayableBySc = 4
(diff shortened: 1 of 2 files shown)


src/smartcontracts/codeMetadata.ts
```diff
@@ -100,13 +119,13 @@ export class CodeMetadata {
     }
 }
 
-enum ByteZero {
+export enum ByteZero {
```
andreibancioiu, line 122: If exported, perhaps we should rename them to be more explicit? E.g. `CodeMetadataByteZero` etc.?

Is it required to export them (this is a low-level detail of `CodeMetadata`)? Or perhaps they are only exported for the sake of the tests?

popenta: I agree with the renaming if they really need to be exported. Maybe some other way of doing was to keep the properties private and have setters and getters, or maybe some methods like `isUpgradeable(): boolean`. Just a thought.

michavie: Good point, seeing it now I dislike it as well. I will propose a new way by placing Byte info in the CodeMetadata namespace

michavie: Updated. Byte info is now only exposed through the CodeMetadata namespace, as in: `CodeMetadata.ByteZero.Upgradeable`

andreibancioiu (approved):

popenta (approved):

PR #354, opened 2023-11-16
Code enhancements

michavie:
- Added code hash computation
- Added `fromHex`
- Added tests

### DIFF
--- a/src/smartcontracts/code.ts
+++ b/src/smartcontracts/code.ts
@@ -16,6 +16,13 @@ export class Code {
     }
 
     /**
+     * Creates a Code object from a hex-encoded string.
+     */
+    static fromHex(hex: string): Code {
+        return new Code(hex)
+    }
+
+    /**
      * Returns the bytecode as a hex-encoded string.
      */
     toString(): string {
--- a/src/smartcontracts/code.ts
+++ b/src/smartcontracts/code.ts
@@ -38,11 +38,11 @@ export class Code {
         return Buffer.from(this.hex, "hex");
     }
 
-    computeHash(): Hash {
+    computeHash(): Buffer {
         const hash = createHasher(CODE_HASH_LENGTH)
             .update(this.valueOf())
-            .digest('hex')
+            .digest();
 
-        return new Hash(hash)
+        return Buffer.from(hash)
     }
 }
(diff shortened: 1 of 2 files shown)


src/smartcontracts/code.ts
```diff
@@ -25,4 +37,12 @@ export class Code {
     valueOf(): Buffer {
         return Buffer.from(this.hex, "hex");
     }
+
+    computeHash(): Hash {
```
andreibancioiu, line 41: It's probable that we'll have some sort of a `computer` object at some point (see the specs for transaction, and hash computation). But for now, it's perfect :+1:

src/smartcontracts/code.ts
```diff
@@ -1,3 +1,8 @@
+import { Hash } from "../hash";
+
+const createHasher = require('blake2b')
```
andreibancioiu, line 3: :+1:

andreibancioiu (dismissed):

src/smartcontracts/code.ts
andreibancioiu, line 41: I think this can directly be a `Buffer`. At some point, we might remove the `Hash` object. E.g. see the sdk-specs for transaction hash: https://github.com/multiversx/mx-sdk-specs/blob/main/sdk-core/transaction.md

src/smartcontracts/code.ts
```diff
@@ -15,6 +20,13 @@ export class Code {
         return new Code(code.toString("hex"));
     }
 
+    /**
+     * Creates a Code object from a hex-encoded string.
+     */
+    static fromHex(hex: string): Code {
```
popenta, line 26: is this really needed? Since it basically is the same as the constructor?

michavie: The constructor is currently private, and it could be made public indeed

Personally, I would still go with this implementation because it hides the internal representation of values, is more readable and provides more control over how code objects are created (e.g. validation added later)

popenta (approved):

src/smartcontracts/code.ts
michavie: Updated

andreibancioiu (approved):

popenta (approved):

PR #359, opened 2023-12-12
Removed old canMint and canBurn arguments

popenta:
The `tokenOperationsFactory` still had `canMint` and `canBurn` that are not needed anymore.
It's more a "fixing change" rather than a breaking change.


andreibancioiu (approved):
Indeed, this should not be considered a breaking change in practice. Therefore, we'll have this change as a minor version :+1:

danielailie (approved):

PR #365, opened 2023-12-19
Set Default Transaction Version=2 and fix failing unit tests

popenta:


CiprianDraghici (approved):

andreibancioiu (approved):
:christmas_tree: 

But we should also handle this in v13:
https://github.com/multiversx/mx-sdk-js-core/issues/367

popenta:
Will do this in the PR that implements the relayed transactions factory.

> 🎄
> 
> But we should also handle this in v13: #367

PR #366, opened 2023-12-22
Implemented the Transaction class

popenta:
Implemented the new `Transaction` class as described in the specs.


src/transaction.spec.ts
```diff
@@ -40,6 +40,28 @@ describe("test transaction construction", async () => {
         assert.deepEqual(transaction.getSignature(), Buffer.from([]));
     });
 
+    it("create transaction from transaction next", async () => {
+        const nextTransaction = new TransactionNext({
```
CiprianDraghici, line 44: I'd extract the transaction payload and use it as an argument to the TransactionNext object and also in each assert.
It would help a lot when you write the asserts to avoid typos, and mismatched values or in case you will extend this test in the future.
This is just a recommendation, nothing wrong with your current implementation.

```
const transactionPayload = {
    sender: "erd1qyu5wthldzr8wx5c9ucg8kjagg0jfs53s8nr3zpz3hypefsdd8ssycr6th",
    receiver: "erd1spyavw0956vq68xj8y4tenjpq2wd5a9p2c6j8gsz7ztyrnpxrruqzu66jx",
    gasLimit: 56000,
    value: "1000000000000000000",
    data: Buffer.from("test"),
    chainID: "T"
}

const nextTransaction = new TransactionNext(transactionPayload);

const transaction = Transaction.fromTransactionNext(nextTransaction);
assert.deepEqual(transaction.getSender(), Address.fromBech32(transactionPayload.sender));
assert.deepEqual(transaction.getReceiver(), Address.fromBech32(transactionPayload.receiver));
assert.equal(transaction.getGasLimit().valueOf(), transactionPayload.gasLimit);
...
```

src/transaction.ts
```diff
@@ -458,3 +492,215 @@ export class TransactionHash extends Hash {
...
+    if (plainTransaction.signature) {
+      delete plainTransaction.signature;
+    }
+
+    if (plainTransaction.guardianSignature) {
+      delete plainTransaction.guardianSignature;
+    }
+
+    if (!plainTransaction.guardian) {
+      delete plainTransaction.guardian
```
CiprianDraghici, line 666: If the `guardian` field is missing, why do you need to try to delete it from `plainTransaction.guardian` ?

src/transactionNext.spec.ts
```diff
@@ -0,0 +1,223 @@
...
+            version: 2,
+            nonce: 89,
+        });
+
+        let serializedTransactionBytes = transactionComputer.computeBytesForSigning(transaction);
+        let serializedTransaction = Buffer.from(serializedTransactionBytes).toString();
+
+        assert.equal(
+            serializedTransaction,
+            `{"nonce":89,"value":"0","receiver":"erd1spyavw0956vq68xj8y4tenjpq2wd5a9p2c6j8gsz7ztyrnpxrruqzu66jx","sender":"erd1qyu5wthldzr8wx5c9ucg8kjagg0jfs53s8nr3zpz3hypefsdd8ssycr6th","gasPrice":1000000000,"gasLimit":50000,"chainID":"D","version":2}`
```
CiprianDraghici, line 49: I would extract the transaction payload

```
const transactionPayload = {
      chainID: networkConfig.ChainID,
      sender: sender,
      receiver: receiver,
      gasLimit: 50000,
      value: 0,
      version: 2,
      nonce: 89,
  }
  let transaction = new TransactionNext(transactionPayload);
// ...
        
```
 and in this asset I'd use: 
 
```
 assert.equal(
  serializedTransaction,
  JSON.stringify(transactionPayload)
);
```

src/transactionNext.spec.ts
```diff
@@ -0,0 +1,223 @@
...
+            nonce: 90,
+            data: new Uint8Array(Buffer.from("hello")),
+        });
+
+        serializedTransactionBytes = transactionComputer.computeBytesForSigning(transaction);
+        serializedTransaction = Buffer.from(serializedTransactionBytes).toString();
+
+        assert.equal(
+            serializedTransaction,
+            `{"nonce":90,"value":"1000000000000000000","receiver":"erd1spyavw0956vq68xj8y4tenjpq2wd5a9p2c6j8gsz7ztyrnpxrruqzu66jx","sender":"erd1qyu5wthldzr8wx5c9ucg8kjagg0jfs53s8nr3zpz3hypefsdd8ssycr6th","gasPrice":1000000000,"gasLimit":70000,"data":"aGVsbG8=","chainID":"D","version":2}`
```
CiprianDraghici, line 68: see previous comment

src/transaction.spec.ts
popenta: Done! Extracted the payload and used it as above. Thanks for the suggestion.

src/transaction.ts
popenta: Because if it's missing it shouldn't be included in the transaction signature. This is how it was also done before. For reference check [the previous implementation](https://github.com/multiversx/mx-sdk-js-core/blob/main/src/transaction.ts#L286).

src/transactionNext.spec.ts
popenta: Well, here I don't think it's that good to do that because the serialized transaction has a specific order and in order to match that with `JSON.stringify()` you'd have to specify the elements in the plain object the same as in the serialized transaction.

src/transactionNext.spec.ts
popenta: same as above

CiprianDraghici (dismissed):

src/transaction.ts
```diff
@@ -435,6 +421,54 @@ export class Transaction {
...
+
+    if (transaction.guardian) {
+      tx.guardian = Address.fromBech32(transaction.guardian)
+    }
+
+    if (transaction.signature.length) {
+      tx.applySignature(transaction.signature);
+    }
+    
+    if (transaction.guardianSignature) {
```
andreibancioiu, line 466: The checks for `transaction.signature` and `transaction.guardianSignature` are not symmetric.

src/transaction.ts
```diff
@@ -458,3 +492,215 @@ export class TransactionHash extends Hash {
   }
 }
 
+/**
+ * An abstraction for creating, signing and broadcasting transactions.
+ * Will replace the {@link Transaction} class in the future.
```
andreibancioiu, line 497: Indeed, this is the one according to sdk-specs :+1:

src/transaction.ts
```diff
@@ -458,3 +492,215 @@ export class TransactionHash extends Hash {
...
+      this.value = value || new BigNumber(0);
+      this.sender = sender;
+      this.receiver = receiver;
+      this.senderUsername = senderUsername || "";
+      this.receiverUsername = receiverUsername || "";
+      this.gasPrice = gasPrice || new BigNumber(TRANSACTION_MIN_GAS_PRICE);
+      this.gasLimit = gasLimit;
+      this.data = data || new Uint8Array();
+      this.chainID = chainID;
+      this.version = version || TRANSACTION_OPTIONS_DEFAULT;
```
andreibancioiu, line 617: Version vs. options.

src/transaction.ts
```diff
@@ -435,6 +421,54 @@ export class Transaction {
...
+  /**
+   * Creates a new Transaction object from a TransactionNext object.
+   */
+  static fromTransactionNext(transaction: ITransactionNext): Transaction {
+    const tx =  new Transaction({
+      sender: Address.fromBech32(transaction.sender),
+      receiver: Address.fromBech32(transaction.receiver),
+      gasLimit: new BigNumber(transaction.gasLimit).toNumber(),
+      chainID: transaction.chainID,
+      value: new BigNumber(transaction.value).toFixed(0),
```
andreibancioiu, line 448: Correct :+1:

src/transaction.ts
```diff
@@ -458,3 +492,215 @@ export class TransactionHash extends Hash {
...
+}
+
+/**
+ * An utilitary class meant to work together with the {@link TransactionNext} class.
+ */
+export class TransactionComputer {
+  constructor() {}
+
+  computeTransactionFee(transaction: ITransactionNext, networkConfig: INetworkConfig): BigNumber{
+    let moveBalanceGas = new BigNumber(
```
andreibancioiu, line 633: When possible, we should use `const` instead of `let` (optional, can also be done in other PR).

src/transaction.ts
```diff
@@ -458,3 +492,215 @@ export class TransactionHash extends Hash {
...
+  computeTransactionHash(transaction: ITransactionNext): Uint8Array{
+    let serializer = new ProtoSerializer();
+
+    const tx = Transaction.fromTransactionNext(transaction);
+    let buffer = serializer.serializeTransaction(tx);
+    let hash = createTransactionHasher(TRANSACTION_HASH_LENGTH)
+      .update(buffer)
+      .digest("hex");
+    
+    return new Uint8Array(Buffer.from(hash, "hex"));
```
andreibancioiu, line 683: Generally speaking, the following:

```
return new Uint8Array(Buffer.from(hash, "hex"))
```

Can be replaced by:

```
return Buffer.from(hash, "hex")
```

(optional, shorter)

src/transaction.ts
```diff
@@ -458,3 +492,215 @@ export class TransactionHash extends Hash {
...
+      sender: transaction.sender,
+      senderUsername: transaction.senderUsername ? Buffer.from(transaction.senderUsername).toString("base64") : undefined,
+      receiverUsername: transaction.receiverUsername ? Buffer.from(transaction.receiverUsername).toString("base64") : undefined,
+      gasPrice: Number(transaction.gasPrice),
+      gasLimit: Number(transaction.gasLimit),
+      data: transaction.data.length == 0 ? undefined : Buffer.from(transaction.data).toString("base64"),
+      chainID: transaction.chainID,
+      version: transaction.version,
+      options: transaction.options == 0 ? undefined : transaction.options,
+      guardian: transaction.guardian == "" ? undefined : transaction.guardian,
```
andreibancioiu, line 700: Check can also be: `transaction.guardian ? undefined : transaction.guardian`, so that it handles `null` values, as well. CC: @CiprianDraghici, makes sense? I think for `options`, as well. Since setters are public.

src/transaction.ts
```diff
@@ -458,3 +492,215 @@ export class TransactionHash extends Hash {
...
+    return {
+      nonce: Number(transaction.nonce),
+      value: new BigNumber(transaction.value).toFixed(0),
+      receiver: transaction.receiver,
+      sender: transaction.sender,
+      senderUsername: transaction.senderUsername ? Buffer.from(transaction.senderUsername).toString("base64") : undefined,
+      receiverUsername: transaction.receiverUsername ? Buffer.from(transaction.receiverUsername).toString("base64") : undefined,
+      gasPrice: Number(transaction.gasPrice),
+      gasLimit: Number(transaction.gasLimit),
+      data: transaction.data.length == 0 ? undefined : Buffer.from(transaction.data).toString("base64"),
```
andreibancioiu, line 696: Since the setter is public, here we can also protect against a null / undefined `data` field.

src/transactionNext.spec.ts
```diff
@@ -0,0 +1,223 @@
...
+import { TransactionNext, TransactionComputer, Transaction } from "./transaction";
+import BigNumber from "bignumber.js";
+import { ProtoSerializer } from "./proto";
+
+class NetworkConfig {
+    MinGasLimit: number;
+    GasPerDataByte: number;
+    GasPriceModifier: number;
+    ChainID: string;
+
```
andreibancioiu, line 12: Now I see that at some point we have to have consistent naming for public fields. Here, uppercase. For `TransactionNext`, the first letter is lowercase.

All right for now, no need to rename the fields of `NetworkConfig` in v13.

src/transaction.ts
popenta: Indeed, they were not. Fixed.

src/transaction.ts
popenta: Fixed

src/transaction.ts
popenta: Replaced with const where possible.

src/transaction.ts
popenta: Indeed, replaced it.

src/transaction.ts
popenta: Check is now:
```js
transaction.guardian ? transaction.guardian : undefined
```

Same for `options`.

src/transaction.ts
popenta: Done!

src/transactionNext.spec.ts
popenta: Not really sure what you mean. The `TransactionComputer` uses the same `INetworkConfig` interface as the `Transaction` class where the first letter is uppercase.

popenta: But I do agree that we need to stay consistent with the naming.

andreibancioiu (approved):

CiprianDraghici (approved):

PR #368, opened 2023-12-31
Add address validator static method without throwing errors

juliancwirko:
Having it as a static method on Address would be super helpful. In many cases, it is good to have a validation without the need to try-catch; at least, I found it helpful. But please remove it if it isn't suitable here or if it should be coded differently. The main goal is to have validation for the address (string) without the need for try-catch.


andreibancioiu (approved):
Thanks for the contribution :pray: 

In the future, this will change a bit, since we'll support addresses with other HRPs, as well.

A general-purpose address validation facility would be added somewhere here (perhaps within the `AddressComputer`):

https://github.com/multiversx/mx-sdk-specs/blob/main/sdk-core/address.md

Additional notes:
 - https://youtu.be/y6Xc35TivRk?si=8tYzXDvcECIY7uyT&t=849

popenta (approved):

Issue #375, opened 2024-02-05
SignableMessage: one does not simply decode a payload that was previously created using "toJSON()"


andreibancioiu:
```
payload = message.toJSON()
// send payload to a backend for verification
signableMessage = ... cumbersome to recover the message from "payload" ...
```

Reported on Telegram. Workaround here: https://t.me/MultiversXDevelopers/123380.

andreibancioiu:
Fixed by https://github.com/multiversx/mx-sdk-js-core/pull/378. Will be released in v13.

PR #378, opened 2024-02-13
Implemented the Message and MessageComputer classes

popenta:
Implemented the `Message` and `MessageComputer` classes based on the [sdk-specs](https://github.com/multiversx/mx-sdk-specs/blob/main/sdk-core/message.md).

### DIFF
--- a/src/interface.ts
+++ b/src/interface.ts
@@ -26,16 +26,36 @@ export interface IPlainTransactionObject {
     guardianSignature?: string;
 }
 
-export interface ISignature { hex(): string; }
-export interface IAddress { bech32(): string; }
-export interface ITransactionValue { toString(): string; }
-export interface IAccountBalance { toString(): string; }
-export interface INonce { valueOf(): number; }
-export interface IChainID { valueOf(): string; }
-export interface IGasLimit { valueOf(): number; }
-export interface IGasPrice { valueOf(): number; }
-export interface ITransactionVersion { valueOf(): number; }
-export interface ITransactionOptions { valueOf(): number; }
+export interface ISignature {
+    hex(): string;
+}
+export interface IAddress {
+    bech32(): string;
+}
+export interface ITransactionValue {
+    toString(): string;
+}
+export interface IAccountBalance {
+    toString(): string;
+}
+export interface INonce {
+    valueOf(): number;
+}
+export interface IChainID {
+    valueOf(): string;
+}
+export interface IGasLimit {
+    valueOf(): number;
+}
+export interface IGasPrice {
+    valueOf(): number;
+}
+export interface ITransactionVersion {
+    valueOf(): number;
+}
+export interface ITransactionOptions {
+    valueOf(): number;
+}
 
 export interface ITransactionPayload {
     length(): number;
@@ -72,4 +92,11 @@ export interface ITransactionNext {
     guardian: string;
     signature: Uint8Array;
     guardianSignature: Uint8Array;
-  }
+}
+
+export interface IMessage {
+    data: Uint8Array;
+    signature?: Uint8Array;
+    address?: IAddress;
+    version?: number;
+}
--- a/src/interface.ts
+++ b/src/interface.ts
@@ -29,30 +29,39 @@ export interface IPlainTransactionObject {
 export interface ISignature {
     hex(): string;
 }
+
 export interface IAddress {
     bech32(): string;
 }
+
 export interface ITransactionValue {
     toString(): string;
 }
+
 export interface IAccountBalance {
     toString(): string;
 }
+
 export interface INonce {
     valueOf(): number;
 }
+
 export interface IChainID {
     valueOf(): string;
 }
+
 export interface IGasLimit {
     valueOf(): number;
 }
+
 export interface IGasPrice {
     valueOf(): number;
 }
+
 export interface ITransactionVersion {
     valueOf(): number;
 }
+
 export interface ITransactionOptions {
     valueOf(): number;
 }
--- a/src/message.spec.ts
+++ b/src/message.spec.ts
@@ -1,0 +1,87 @@
+import { assert } from "chai";
+import { Message, MessageComputer } from "./message";
+import { loadTestWallets, TestWallet } from "./testutils";
+import { UserVerifier } from "@multiversx/sdk-wallet";
+import { DEFAULT_MESSAGE_VERSION } from "./constants";
+
+describe("test message", () => {
+    let alice: TestWallet;
+    const messageComputer = new MessageComputer();
+
+    before(async function () {
+        ({ alice } = await loadTestWallets());
+    });
+
+    it("should test message compute bytes for signing", async () => {
+        const data = Buffer.from("test message");
+
+        const message = new Message({
+            data: data,
+        });
+
+        const serialized = messageComputer.computeBytesForSigning(message);
+
+        assert.equal(
+            Buffer.from(serialized).toString("hex"),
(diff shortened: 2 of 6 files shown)


src/message.ts
```diff
@@ -0,0 +1,93 @@
...
+     */
+    public signature?: Uint8Array;
+    /**
+     * Address of the wallet that performed the signing operation.
+     */
+    public address?: IAddress;
+    /**
+     * Number representing the message version.
+     */
+    public version?: number;
```
andreibancioiu, line 23: Can be non-optional.

src/message.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    constructor(options: { data: Uint8Array; signature?: Uint8Array; address?: IAddress; version?: number }) {
+        this.data = options.data;
+        this.signature = options.signature;
+        this.address = options.address;
+        this.version = options.version ? options.version : DEFAULT_MESSAGE_VERSION;
```
andreibancioiu, line 29: Somehow unexpected when version will be set to `0` (but all right in practice, since there is no version `0`). Maybe check for `undefined` instead of `options.version`?

(optional)

src/message.ts
```diff
@@ -0,0 +1,93 @@
...
+        return {
+            message: Buffer.from(message.data).toString("hex"),
+            signature: message.signature ? Buffer.from(message.signature).toString("hex") : "",
+            address: message.address ? message.address.bech32() : "",
+            version: message.version ? message.version : DEFAULT_MESSAGE_VERSION,
+        };
+    }
+
+    unpackMessage(packedMessage: { message: string; signature?: string; address?: string; version?: number }): Message {
+        let dataHex = this.trimHexPrefix(packedMessage.message);
```
andreibancioiu, line 63: `const` if possible.

src/message.ts
```diff
@@ -0,0 +1,93 @@
...
+        const data = Buffer.from(dataHex, "hex");
+
+        let signature: Uint8Array | undefined = undefined;
+        if (packedMessage.signature) {
+            let signatureHex = this.trimHexPrefix(packedMessage.signature);
+            signature = Buffer.from(signatureHex, "hex");
+        }
```
andreibancioiu, line 70: Can also be (proposal):

```
const signatureHex = this.trimHexPrefix(packedMessage.signature || "");
const signature = Buffer.from(signatureHex, "hex")
```

Will be an empty buffer if signature not provided, but that is maybe even better.

src/transactionsFactories/relayedTransactionsFactory.ts
```diff
@@ -13,6 +13,9 @@ interface IConfig {
     gasLimitPerByte: BigNumber.Value;
 }
 
+/**
+ * Use this class to create both RelyedV1 and RelayedV2 transactions.
```
andreibancioiu, line 17: Typo.

src/interface.ts
```diff
@@ -72,4 +101,11 @@ export interface ITransactionNext {
     guardian: string;
     signature: Uint8Array;
     guardianSignature: Uint8Array;
-  }
+}
+
+export interface IMessage {
```
andreibancioiu, line 106: Maybe here we can live without the interface (related to the brainstorming on sdk-py).

Thus, maybe the DTO should suffice, for now? Then introduce an interface when needed? CC: @ccorcoveanu, as well.

src/message.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+
+        assert.equal(
+            Buffer.from(message.signature).toString("hex"),
+            "7aff43cd6e3d880a65033bf0a1b16274854fd7dfa9fe5faa7fa9a665ee851afd4c449310f5f1697d348e42d1819eaef69080e33e7652d7393521ed50d7427a0e",
+        );
+
+        const packedMessage = messageComputer.packMessage(message);
+        assert.deepEqual(packedMessage, {
+            address: "erd1qyu5wthldzr8wx5c9ucg8kjagg0jfs53s8nr3zpz3hypefsdd8ssycr6th",
+            message: "74657374",
```
andreibancioiu, line 48: And now, re-re-thinking about it :see_no_evil: 

@ccorcoveanu, all right if we do it this way? That is, without the `0x` prefix. Seems less cumbersome. And when people see `signature` they automatically think: _well, that is hex, of course._ And `message` is the same, for consistency. How do you feel about it?

For context: https://github.com/multiversx/mx-sdk-js-core/issues/375

src/signableMessage.ts
```diff
@@ -1,80 +1,81 @@
...
+
+    serializeForSigning(): Buffer {
+        const messageSize = Buffer.from(this.message.length.toString());
+        const signableMessage = Buffer.concat([messageSize, this.message]);
+        let bytesToHash = Buffer.concat([Buffer.from(MESSAGE_PREFIX), signableMessage]);
+
+        return createKeccakHash("keccak256").update(bytesToHash).digest();
+    }
+
+    serializeForSigningRaw(): Buffer {
```
andreibancioiu, line 53: I wonder where is this used.

andreibancioiu (commented):
:rocket:

src/message.ts
popenta: Done

src/message.ts
popenta: As discussed, since `version = 0` it's not valid it ok how it is.

src/message.ts
popenta: changed to `const`

src/message.ts
popenta: Replaced with your proposal. Thanks!

src/transactionsFactories/relayedTransactionsFactory.ts
popenta: Fixed

src/signableMessage.ts
popenta: Apparently it's not used anywhere. Should I delete it?

ccorcoveanu: https://github.com/multiversx/mx-sdk-js-hw-provider/blob/b6dcf0c7863f3ad6fe157163e0227e71d69bf9f3/src/hwProvider.ts#L346

src/message.ts
ccorcoveanu: `this.version = options.version || DEFAULT_MESSAGE_VERSION` would be shorter
(optional)

src/message.spec.ts
ccorcoveanu: I like the no-0x version better

src/message.ts
popenta: Changed

src/interface.ts
popenta: Removed the interface.

src/signableMessage.ts
popenta: Will not delete it.

src/transaction.ts
```diff
@@ -653,7 +653,7 @@ export class TransactionComputer {
 
     const serialized = JSON.stringify(plainTransaction);
 
-    return new Uint8Array(Buffer.from(serialized));
+    return Buffer.from(serialized);
```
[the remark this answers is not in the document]
popenta: Well, in theory it would break compatibility, but this has not been yet released so there can't be any apps that use it. And since `Buffer` implements `Uint8Array` I assume it's all right.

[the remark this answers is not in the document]
andreibancioiu: Indeed, `TransactionComputer` is new in v13.

andreibancioiu (approved):

ccorcoveanu (approved):

Issue #379, opened 2024-02-15
Send NFT Functionality Not Working in MultiversX SDK-Core


karthik-durai:
Upon implementing the send NFT functionality using MultiversX SDK-Core, it's observed that the sender and destination addresses in the transaction call are identical, leading to unsuccessful transfers.

Steps to Reproduce:
1. create `tokenTransfer` object using `nonFungible` function from `TokenTransfer` with token identifier and nonce.
2. create transaction object using `createESDTNFTTransfer` by passing
   a. tokenTransfer
   b. nonce
   c. sender address
   d. destination address
   e. chain id
3. sign the transaction by passing transaction object.
4. broadcast transaction.
5. Inspect the transaction call parameters in the network tab of the browser and also transaction explorer of multiversx.
6. Notice that sender and destination addresses are the same.

andreibancioiu:
Hello @karthik-durai,

On MultiversX, for NFT transfers, `transaction.sender == transaction.receiver`, while the actual destination address is encoded in the transaction payload.

E.g.

```
NFT transfer transaction = {
    Sender: <account address of the sender>
    Receiver: <same as sender>
    Data: "ESDTNFTTransfer ... "
   ...
}
```

Additional resources:
 - https://docs.multiversx.com/tokens/nft-tokens/#transfers
 - https://devnet-explorer.multiversx.com/transactions?function=ESDTNFTTransfer

karthik-durai:
thanks @andreibancioiu , close this PR.

PR #380, opened 2024-02-15
[FIX] Issue - 379 - unable to send NFT to any destination address

karthik-durai:
#379 

Changes:
1. Replaced sender address with destination address prop.


karthik-durai:
not needed

ref: https://github.com/multiversx/mx-sdk-js-core/issues/379

PR #382, opened 2024-02-15
Commute to ES2020. Partially replace "BigNumber.Value" with "bigint"

andreibancioiu:
- `compilerOptions`: for `target`, use ES2020 instead of ES2015.
 - partially replace `BigNumber.Value` with `bigint` (on some components that are new in v13)

See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt

### DIFF
--- a/src/tokens.spec.ts
+++ b/src/tokens.spec.ts
@@ -1,12 +1,12 @@
-import { Token, TokenComputer } from "./tokens";
 import { assert } from "chai";
+import { Token, TokenComputer } from "./tokens";
 
 describe("test token computer", async () => {
     const tokenComputer = new TokenComputer();
 
     it("should test if token is fungible", async () => {
-        const fungibleToken = new Token("TEST-123456", 0);
-        const nonFungibleToken = new Token("NFT-987654", 7);
+        const fungibleToken = new Token("TEST-123456", 0n);
+        const nonFungibleToken = new Token("NFT-987654", 7n);
 
         assert.equal(tokenComputer.isFungible(fungibleToken), true);
         assert.equal(tokenComputer.isFungible(nonFungibleToken), false);
--- a/src/tokens.ts
+++ b/src/tokens.ts
@@ -1,11 +1,10 @@
-import BigNumber from "bignumber.js";
 import { ErrInvalidTokenIdentifier } from "./errors";
 
 export class Token {
     identifier: string;
-    nonce: BigNumber.Value;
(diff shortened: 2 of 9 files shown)


src/tokens.spec.ts
```diff
@@ -1,12 +1,12 @@
...
 import { assert } from "chai";
+import { Token, TokenComputer } from "./tokens";
 
 describe("test token computer", async () => {
     const tokenComputer = new TokenComputer();
 
     it("should test if token is fungible", async () => {
-        const fungibleToken = new Token("TEST-123456", 0);
-        const nonFungibleToken = new Token("NFT-987654", 7);
+        const fungibleToken = new Token("TEST-123456", 0n);
```
popenta, line 8: I'd say it's more explicit to call `BigInt()` than to add `n` at the end of the value, but it's easier and shorter this way.

src/tokens.ts
```diff
@@ -25,7 +24,7 @@ export class TokenComputer {
     constructor() {}
 
     isFungible(token: Token): boolean {
-        return token.nonce === 0;
+        return token.nonce == 0n;
```
popenta, line 27: Is there any reason to not use `===`?

src/transactionsFactories/smartContractTransactionsFactory.spec.ts
```diff
@@ -1,15 +1,15 @@
+import BigNumber from "bignumber.js";
```
popenta, line 1: I don't think this is used anymore. Can be deleted if so.

andreibancioiu: Still needed on `smartContractFactory.createTransactionForExecute`, which is not yet adjusted to use `bigint` (to be done in another PR).

src/tokens.ts
andreibancioiu: Helps a bit if someone uses sdk-core from pure JavaScript, and aren't constrained (by TS) to pass a `bigint` instead of a `number`. E.g.

```
> BigInt(0) == 0
true
> BigInt(0) === 0
false
```

But fixed, using `===` now.

src/tokens.spec.ts
andreibancioiu: As discussed separately, we can recommend usage of bigint literal (*n).

ccorcoveanu (approved):

PR #386, opened 2024-02-20
Refactor Transaction Watcher

popenta:
**Later edit:** the breaking part was recovered in https://github.com/multiversx/mx-sdk-js-core/pull/404.

~~This PR introduces **breaking changes**.~~

Before, the methods in the `TransactionWatcher` class required an input parameter of type `ITransaction`. Now, in order to be compatible with the `TransactionNext` class, the methods simply require the transaction hash.

Migrated from:
```js
public async awaitCompleted(transaction: ITransaction): Promise<ITransactionOnNetwork> {...}
```
to:
```js
public async awaitCompleted(txHash: string): Promise<ITransactionOnNetwork> {...}
```

### DIFF
--- a/src/transaction.local.net.spec.ts
+++ b/src/transaction.local.net.spec.ts
@@ -142,23 +142,23 @@ describe("test transaction", function () {
     it("should create transaction using the NextTokenTransferFactory", async function () {
         this.timeout(70000);
 
-        let provider = createLocalnetProvider();
-        let watcher = new TransactionWatcher({
+        const provider = createLocalnetProvider();
+        const watcher = new TransactionWatcher({
             getTransaction: async (hash: string) => {
                 return await provider.getTransaction(hash, true);
             },
         });
 
-        let network = await provider.getNetworkConfig();
+        const network = await provider.getNetworkConfig();
 
         const config = new TransactionsFactoryConfig(network.ChainID);
         const factory = new NextTransferTransactionsFactory(config, new TokenComputer());
 
         await alice.sync(provider);
         await bob.sync(provider);
-        let initialBalanceOfBob = new BigNumber(bob.account.balance.toString());
+        const initialBalanceOfBob = new BigNumber(bob.account.balance.toString());
 
-        let transaction = factory.createTransactionForNativeTokenTransfer({
+        const transaction = factory.createTransactionForNativeTokenTransfer({
             sender: alice.address,
             receiver: bob.address,
             nativeAmount: 42000000000000000000n,
@@ -174,7 +174,7 @@ describe("test transaction", function () {
         await watcher.awaitCompleted(txHash);
 
         await bob.sync(provider);
-        let newBalanceOfBob = new BigNumber(bob.account.balance.toString());
+        const newBalanceOfBob = new BigNumber(bob.account.balance.toString());
 
         assert.deepEqual(TokenTransfer.egldFromAmount(42).valueOf(), newBalanceOfBob.minus(initialBalanceOfBob));
     });
--- a/src/transactionsFactories/transferTransactionsFactory.ts
+++ b/src/transactionsFactories/transferTransactionsFactory.ts
@@ -23,8 +23,8 @@ interface TokenComputer {
 
 /**
  * Use this class to create transactions for native token transfers (EGLD) or custom tokens transfers (ESDT/NTF/MetaESDT).
+ * This name is only temporary, the class will be renamed to `TransferTransactionsFactory`.
  */
-// this name is only temporary; the class will be renamed to `TransferTransactionsFactory`
 export class NextTransferTransactionsFactory {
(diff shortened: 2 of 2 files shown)


src/smartcontracts/smartContractResults.local.net.spec.ts
```diff
@@ -58,8 +58,8 @@ describe("fetch transactions from local testnet", function () {
         await provider.sendTransaction(transactionDeploy);
         await provider.sendTransaction(transactionIncrement);
 
-        await watcher.awaitCompleted(transactionDeploy);
-        await watcher.awaitCompleted(transactionIncrement);
+        await watcher.awaitCompleted(transactionDeploy.getHash().hex());
```
andreibancioiu, line 61: In a future PR, we should also adjust the integration tests to use the transaction factories for deploy, call etc.

src/transaction.local.net.spec.ts
```diff
@@ -132,7 +139,47 @@ describe("test transaction", function () {
         Logger.trace(JSON.stringify(await provider.simulateTransaction(transactionTwo), null, 4));
     });
 
-    async function signTransaction(options: { transaction: Transaction, wallet: TestWallet }) {
+    it("should create transaction using the NextTokenTransferFactory", async function () {
+        this.timeout(70000);
+
+        let provider = createLocalnetProvider();
```
andreibancioiu, line 145: Here & below, use `const` when applicable.

src/smartcontracts/smartContract.local.net.spec.ts
```diff
@@ -325,4 +330,51 @@ describe("test on local testnet", function () {
         queryResponse = await provider.queryContract(query);
         assert.equal(decodeUnsignedNumber(queryResponse.getReturnDataParts()[0]), 0);
     });
+
+    it("counter: should deploy and call using the SmartContractFactory", async function () {
```
andreibancioiu, line 334: :rocket:  :+1: (in relation to the other comment about a next PR for integration tests, only now I see this test)

src/smartcontracts/smartContract.local.net.spec.ts
```diff
@@ -325,4 +330,51 @@ describe("test on local testnet", function () {
...
+        this.timeout(80000);
+
+        TransactionWatcher.DefaultPollingInterval = 5000;
+        TransactionWatcher.DefaultTimeout = 50000;
+
+        const network = await provider.getNetworkConfig();
+        await alice.sync(provider);
+
+        const transactionComputer = new TransactionComputer();
+        const config = new TransactionsFactoryConfig(network.ChainID);
```
andreibancioiu, line 344: Now I see - we should pass init parameters to `TransactionsFactoryConfig` as options (even if now there's only one, the chain ID) - similar to `SmartContractTransactionsFactory` etc. This, in order to avoid breaking changes in the future, if the config should be extended to receive an extra parameter in the constructor.

This would be for another, separate PR.

src/smartcontracts/smartContract.local.net.spec.ts
```diff
@@ -325,4 +330,51 @@ describe("test on local testnet", function () {
...
+        const factory = new SmartContractTransactionsFactory({ config: config, tokenComputer: new TokenComputer() });
+
+        let bytecode = await promises.readFile("src/testdata/counter.wasm");
+
+        const deployTransaction = factory.createTransactionForDeploy({
+            sender: alice.address,
+            bytecode: bytecode,
+            gasLimit: 3000000n,
+        });
+        deployTransaction.nonce = BigInt(alice.account.nonce.valueOf());
```
andreibancioiu, line 354: A bit verbose, but good at this moment. In the cookbook I think we will not mention `AccountNonce` class anymore - left for the client code to hold & handle the nonce.

andreibancioiu (commented):
Nice :rocket: 

We should also add the prefix "Breaking change" in the PR title, even if the note exists already in the description.

src/transactionsFactories/transferTransactionsFactory.ts
```diff
@@ -24,6 +24,7 @@ interface TokenComputer {
 /**
  * Use this class to create transactions for native token transfers (EGLD) or custom tokens transfers (ESDT/NTF/MetaESDT).
  */
+// this name is only temporary; the class will be renamed to `TransferTransactionsFactory`
```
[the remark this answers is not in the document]
popenta: Moved it.

src/transaction.local.net.spec.ts
popenta: Done!

andreibancioiu (approved):

bogdan-rosianu (approved):

andreibancioiu:
The breaking part was recovered in https://github.com/multiversx/mx-sdk-js-core/pull/404.

PR #387, opened 2024-02-21
Changed the constructor of TransactionFactoryConfig

popenta:
The constructor of `TransactionFactoryConfig` used to need only one parameter, `chainId: string`. Now, it should receive a plain object of type:
```js
options: { chainID: string }
```

The name also changed, from `chainId` to `chainID` to be consistent since the member of the class is named `chainID`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@multiversx/sdk-core",
-  "version": "13.0.0-beta.2",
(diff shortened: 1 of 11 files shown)


andreibancioiu (approved):
We can say that it's not a breaking change (feature is new in v13).

bogdan-rosianu (approved):

PR #389, opened 2024-02-23
Extended localnet tests to use the transaction factories, as well

popenta:
Duplicated the tests that run on `localnet` to use the transaction factories. The old tests will be removed in the future, when the functionality will be deleted.


CiprianDraghici (dismissed):

CiprianDraghici (approved):

src/testutils/contractController.ts
```diff
@@ -55,4 +58,10 @@ export class ContractController {
         Logger.debug(`ContractController.query [end]: function = ${interaction.getFunction()}, return code = ${bundle.returnCode}`);
         return bundle;
     }
+
+    private getTransactionHash(transaction: Transaction): string {
```
andreibancioiu, line 62: Generally speaking, `hash` should also have been returned by `await this.provider.sendTransaction(transaction);`.

src/transaction.local.net.spec.ts
```diff
@@ -143,11 +148,14 @@ describe("test transaction", function () {
...
-            getTransaction: async (hash: string) => {
-                return await provider.getTransaction(hash, true);
+        const watcher = new TransactionWatcher(
+            {
+                getTransaction: async (hash: string) => {
+                    return await provider.getTransaction(hash, true);
+                },
             },
-        });
+            { timeoutMilliseconds: 100000 },
```
andreibancioiu, line 157: Could have added a utility test function such as `createTransactionWatcher`.

andreibancioiu (approved):

src/transaction.local.net.spec.ts
popenta: added

src/testutils/contractController.ts
popenta: done

andreibancioiu (approved):

CiprianDraghici (approved):

PR #395, opened 2024-03-06
Adaptor for executing smart contract queries

popenta:


src/smartContractQueriesController.ts
```diff
@@ -21,14 +20,19 @@ interface ILegacyQuery {
     getEncodedArguments(): string[];
 }
 
+interface IQueryRunner {
+    networkProvider: INetworkProvider;
```
andreibancioiu, line 24: This should not be needed. Generally speaking, now, the queries controller should not know about any network provider.

src/adapters/smartContractQueriesAdapter.ts
```diff
@@ -0,0 +1,44 @@
...
+interface ILegacyQuery {
+    address: IAddress;
+    caller?: IAddress;
+    func: { toString(): string };
+    value?: { toString(): string };
+    getEncodedArguments(): string[];
+}
+
+export class SmartContractQueriesAdapter {
+    networkProvider: INetworkProvider;
```
andreibancioiu, line 19: Can be private, readonly.

src/adapters/smartContractQueriesAdapter.ts
```diff
@@ -0,0 +1,44 @@
+import { Address } from "../address";
+import { IAddress } from "../interface";
+import { IContractQueryResponse } from "../interfaceOfNetwork";
+import { SmartContractQuery, SmartContractQueryResponse } from "../smartContractQuery";
+
+interface INetworkProvider {
+    queryContract(query: ILegacyQuery): Promise<IContractQueryResponse>;
```
andreibancioiu, line 7: Indeed.

src/adapters/smartContractQueriesAdapter.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export class SmartContractQueriesAdapter {
+    networkProvider: INetworkProvider;
+
+    constructor(options: { networkProvider: INetworkProvider }) {
+        this.networkProvider = options.networkProvider;
+    }
+
+    async queryContract(query: SmartContractQuery): Promise<SmartContractQueryResponse> {
```
andreibancioiu, line 25: :+1:

src/adapters/smartContractQueriesAdapter.ts
```diff
@@ -0,0 +1,44 @@
...
+
+interface ILegacyQuery {
+    address: IAddress;
+    caller?: IAddress;
+    func: { toString(): string };
+    value?: { toString(): string };
+    getEncodedArguments(): string[];
+}
+
+export class SmartContractQueriesAdapter {
```
andreibancioiu, line 18: Let's think of a general rule for naming the classes in the `adapters` package.

E.g., this can be, I think, `NetworkProviderAsQueryRunnerAdapter`.

And the upcoming adapter required by `TransactionWatcher` would be: `ProxyNetworkProviderAsTransactionFetcherAdapter`.

How does this sound? We should ask for some opinions.

src/smartContractQueriesController.ts
```diff
@@ -21,14 +20,19 @@ interface ILegacyQuery {
     getEncodedArguments(): string[];
 }
 
+interface IQueryRunner {
```
andreibancioiu, line 23: :+1:

src/smartContractQueriesController.ts
popenta: Indeed, removed

src/adapters/smartContractQueriesAdapter.ts
popenta: made it private, readonly

src/adapters/smartContractQueriesAdapter.ts
danielailie: From what I see here the query method actually returns a SmartContractQueryResponse and it is the only method so, I would keep the SmartContractQueriesAdapter

src/adapters/smartContractQueriesAdapter.ts
```diff
@@ -0,0 +1,44 @@
+import { Address } from "../address";
+import { IAddress } from "../interface";
+import { IContractQueryResponse } from "../interfaceOfNetwork";
+import { SmartContractQuery, SmartContractQueryResponse } from "../smartContractQuery";
+
+interface INetworkProvider {
+    queryContract(query: ILegacyQuery): Promise<IContractQueryResponse>;
+}
+
```
danielailie, line 9: Why are the interfaces in this file, it doesn't look like they are in the right place. The names don't have any connection

andreibancioiu (approved):

src/adapters/queryRunnerAdapter.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    async runQuery(query: SmartContractQuery): Promise<SmartContractQueryResponse> {
+        const legacyQuery: IQuery = {
+            address: Address.fromBech32(query.contract),
+            caller: query.caller ? Address.fromBech32(query.caller) : undefined,
+            func: query.function,
+            value: query.value,
+            getEncodedArguments: () => query.arguments.map((arg) => Buffer.from(arg).toString("hex")),
+        };
+
```
danielailie, line 33: can you also update the name to this?

danielailie (approved):

PR #398, opened 2024-03-12
Re-write type expression parser

andreibancioiu:
Some type expressions couldn't be parsed before. E.g.
 - `variadic<multi<BigUint,BigUint,u64,BigUint>>`
 - `multi<u8, utf-8 string, u8, utf-8 string, u8, utf-8 string>`

Changes:
 - Added `TypeFormula` and `TypeFormulaParser` under a new package: `abi`. The classes in this package will be kept "under the radar" (not listed in the default exports) for some time. In the future, e.g. v14, they will replace the legacy `Type` and `TypeExpressionParser`.
 - Adjusted existing `TypeExpressionParser`: removed tricky JSON-based parsing; using the new `typeFormulaParser` as a "backend" for parsing type expressions.
 - Removed dependency `json-duplicate-key-handle`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@multiversx/sdk-core",
-  "version": "13.0.0-beta.4",
+  "version": "13.0.0-beta.5",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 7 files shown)


danielailie (approved):
:+1:

radumojic (approved):

PR #399, opened 2024-03-13
Inherit version and options from inner transaction when building relayedV2 transaction

popenta:


src/relayedTransactionV2Builder.ts
```diff
@@ -98,6 +98,8 @@ export class RelayedTransactionV2Builder {
                 this.innerTransactionGasLimit.valueOf() + this.netConfig.MinGasLimit + this.netConfig.GasPerDataByte * payload.length(),
             data: payload,
             chainID: this.netConfig.ChainID,
+            version: this.innerTransaction.getVersion(),
```
andreibancioiu, line 101: Good, for relayed V2 we use whatever the inner transaction has.

andreibancioiu (approved):

radumojic (approved):

andreibancioiu (approved):

PR #406, opened 2024-03-20
Transaction signing by hash

popenta:
If the `version` and `options` field are set accordingly, the transaction will be signed by hash.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -16,4 +16,5 @@ export const DEFAULT_HRP = "erd";
 export const ESDT_CONTRACT_ADDRESS = "erd1qqqqqqqqqqqqqqqpqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqzllls8a5w6u";
 export const DEFAULT_MESSAGE_VERSION = 1;
 export const MESSAGE_PREFIX = "\x17Elrond Signed Message:\n";
-export const DEFAULT_HEX_HASH_LENGTH = 64;
+export const HEX_TRANSACTION_HASH_LENGTH = 64;
+export const BECH32_ADDRESS_LENGTH = 62;
--- a/src/transaction.spec.ts
+++ b/src/transaction.spec.ts
@@ -614,4 +614,43 @@ describe("test transaction", async () => {
         transaction.applyGuardianSignature(transaction.getSignature());
         assert.isTrue(transaction.isGuardedTransaction());
     });
+
+    it("test sign using hash", async () => {
+        let transaction = new Transaction({
+            nonce: 89n,
+            value: 0n,
+            sender: wallets.alice.address.toBech32(),
+            receiver: wallets.bob.address.toBech32(),
+            gasLimit: 50000n,
+            chainID: "integration tests chain ID",
+            version: 2,
+            options: 1,
+        });
+
+        transaction.signature = await wallets.alice.signer.sign(
+            Buffer.from(transactionComputer.computeBytesForSigning(transaction)),
+        );
+
+        assert.equal(
+            "f0c81f2393b1ec5972c813f817bae8daa00ade91c6f75ea604ab6a4d2797aca4378d783023ff98f1a02717fe4f24240cdfba0b674ee9abb18042203d713bc70a",
+            Buffer.from(transaction.signature).toString("hex"),
+        );
+    });
+
+    it("should apply guardian", async () => {
+        let transaction = new Transaction({
+            nonce: 89n,
+            value: 0n,
+            sender: wallets.alice.address.toBech32(),
+            receiver: wallets.bob.address.toBech32(),
+            gasLimit: 50000n,
+            chainID: "localnet",
+        });
+
+        transactionComputer.applyGuardian(transaction, wallets.carol.address.toBech32());
+
+        assert.equal(transaction.version, 2);
+        assert.equal(transaction.options, 2);
+        assert.equal(transaction.guardian, wallets.carol.address.toBech32());
+    });
 });
--- a/src/transaction.spec.ts
+++ b/src/transaction.spec.ts
@@ -622,6 +622,7 @@ describe("test transaction", async () => {
             sender: wallets.alice.address.toBech32(),
             receiver: wallets.bob.address.toBech32(),
             gasLimit: 50000n,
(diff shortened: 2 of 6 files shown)


src/transaction.spec.ts
```diff
@@ -614,4 +614,43 @@ describe("test transaction", async () => {
...
+            version: 2,
+            options: 1,
+        });
+
+        transaction.signature = await wallets.alice.signer.sign(
+            Buffer.from(transactionComputer.computeBytesForSigning(transaction)),
+        );
+
+        assert.equal(
+            "f0c81f2393b1ec5972c813f817bae8daa00ade91c6f75ea604ab6a4d2797aca4378d783023ff98f1a02717fe4f24240cdfba0b674ee9abb18042203d713bc70a",
```
bogdan-rosianu, line 636: validated on mx-chain-go's integration tests 👍

src/transactionComputer.ts
```diff
@@ -79,5 +104,17 @@ export class TransactionComputer {
         return value && value.length ? Buffer.from(value).toString("base64") : undefined;
     }
 
-    // TODO: add missing functions wrt. specs
+    private ensureFields(transaction: ITransaction) {
+        if (transaction.sender.length !== BECH32_ADDRESS_LENGTH) {
+            throw new errors.ErrBadUsage("The `sender` field is not set");
```
bogdan-rosianu, line 109: in case of both sender and receiver, the error is wrong. the field might be set, but invalid. for chain ID, the error message is ok

popenta: changed the error message

bogdan-rosianu (dismissed):

bogdan-rosianu (approved):

src/transaction.spec.ts
andreibancioiu: Thanks :pray:

src/transaction.spec.ts
```diff
@@ -614,4 +614,44 @@ describe("test transaction", async () => {
...
+        let transaction = new Transaction({
+            nonce: 89n,
+            value: 0n,
+            sender: wallets.alice.address.toBech32(),
+            receiver: wallets.bob.address.toBech32(),
+            gasLimit: 50000n,
+            gasPrice: 1000000000n,
+            chainID: "integration tests chain ID",
+            version: 2,
+            options: 1,
```
andreibancioiu, line 628: In the specs, we should add an extra method on the computer - e.g. _apply option sign using hash._

src/transaction.spec.ts
```diff
@@ -614,4 +614,44 @@ describe("test transaction", async () => {
...
+        let transaction = new Transaction({
+            nonce: 89n,
+            value: 0n,
+            sender: wallets.alice.address.toBech32(),
+            receiver: wallets.bob.address.toBech32(),
+            gasLimit: 50000n,
+            chainID: "localnet",
+        });
+
+        transactionComputer.applyGuardian(transaction, wallets.carol.address.toBech32());
```
andreibancioiu, line 651: :rocket:

src/transactionComputer.ts
```diff
@@ -57,6 +68,20 @@ export class TransactionComputer {
...
+    hasOptionsSetForHashSigning(transaction: ITransaction): boolean {
+        return (transaction.options & TRANSACTION_OPTIONS_TX_HASH_SIGN) == TRANSACTION_OPTIONS_TX_HASH_SIGN;
+    }
+
+    applyGuardian(transaction: ITransaction, guardian: string) {
+        transaction.version = 2;
+        transaction.options = transaction.options | TRANSACTION_OPTIONS_TX_GUARDED;
+        transaction.guardian = guardian;
+    }
+
```
andreibancioiu, line 84: Nice! We should also think of an `apply*` function for setting `TRANSACTION_OPTIONS_TX_HASH_SIGN`.

src/transactionComputer.ts
```diff
@@ -41,14 +43,23 @@ export class TransactionComputer {
     }
 
     computeBytesForSigning(transaction: ITransaction): Uint8Array {
-        // TODO: do some checks for the transaction e.g. sender, chain ID etc.
-        // TODO: for appropriate tx.version, interpret tx.options accordingly and sign using the content / data hash
+        this.ensureFields(transaction);
```
andreibancioiu, line 46: We can call this `validate()` or `checkValidity()` or `ensureValid()`. And then, within it, we can also check the following constraint: `if tx.version is 1 but an option is set, raise error`.

Can be done in a separate PR.

src/transactionComputer.ts
```diff
@@ -79,5 +104,17 @@ export class TransactionComputer {
         return value && value.length ? Buffer.from(value).toString("base64") : undefined;
     }
 
-    // TODO: add missing functions wrt. specs
+    private ensureFields(transaction: ITransaction) {
```
andreibancioiu, line 107: Can be a function that covers more constraints on the purpose of validating the transaction prior signing. E.g. options vs. version. Can be done in another PR.

andreibancioiu (approved):

src/constants.ts
```diff
@@ -16,4 +16,5 @@ export const DEFAULT_HRP = "erd";
 export const ESDT_CONTRACT_ADDRESS = "erd1qqqqqqqqqqqqqqqpqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqzllls8a5w6u";
 export const DEFAULT_MESSAGE_VERSION = 1;
 export const MESSAGE_PREFIX = "\x17Elrond Signed Message:\n";
-export const DEFAULT_HEX_HASH_LENGTH = 64;
+export const HEX_TRANSACTION_HASH_LENGTH = 64;
```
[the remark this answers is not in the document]
popenta: indeed

src/transaction.spec.ts
popenta: Created a task, will do.

src/transactionComputer.ts
popenta: Will do in a future PR.

src/transactionComputer.ts
popenta: Will be done in the next PR.

PR #411, opened 2024-03-22
Addeed a method to apply options for hash signing

popenta:


src/transactionComputer.ts
```diff
@@ -77,11 +82,21 @@ export class TransactionComputer {
     }
 
     applyGuardian(transaction: ITransaction, guardian: string) {
-        transaction.version = 2;
+        if (transaction.version < TRANSACTION_VERSION_DEFAULT) {
+            transaction.version = TRANSACTION_VERSION_DEFAULT;
+        }
```
andreibancioiu, line 87: Tricky, but correct :+1: 

Since `TRANSACTION_VERSION_DEFAULT >= 2`.

andreibancioiu: A separate constant called `TRANSACTION_MIN_VERSION_THAT_SUPPORTS_OPTIONS` would be more explicit. This is optional, however, just an idea / thinking out loud.

src/transactionComputer.ts
```diff
@@ -116,5 +131,15 @@ export class TransactionComputer {
         if (!transaction.chainID.length) {
             throw new errors.ErrBadUsage("The `chainID` field is not set");
         }
+
+        if (transaction.version < TRANSACTION_VERSION_DEFAULT) {
```
andreibancioiu, line 135: Tricky if we will have, at some point, a default of 3.

`TRANSACTION_MIN_VERSION_THAT_SUPPORTS_OPTIONS` might come in handy here (just an idea).

src/transactionComputer.ts
```diff
@@ -116,5 +131,15 @@ export class TransactionComputer {
...
+            if (this.hasOptionsSetForGuardedTransaction(transaction) || this.hasOptionsSetForHashSigning(transaction)) {
+                throw new errors.ErrBadUsage(
+                    "The `version` field should be " +
```
andreibancioiu, line 138: `Non-empty transaction options require transaction version >= ${TRANSACTION_MIN_VERSION_THAT_SUPPORTS_OPTIONS}`.

src/transaction.spec.ts
```diff
@@ -646,12 +639,63 @@ describe("test transaction", async () => {
             receiver: wallets.bob.address.toBech32(),
             gasLimit: 50000n,
             chainID: "localnet",
+            options: 1,
```
andreibancioiu, line 642: Previous value (missing) works as well, correct?

Perhaps keep the test as it was, but duplicate it as "it(should apply guardian (when using hash signing))"?

src/transaction.spec.ts
```diff
@@ -204,16 +204,9 @@ describe("test transaction", async () => {
...
-            "c83e69b853a891bf2130c1839362fe2a7a8db327dcc0c9f130497a4f24b0236140b394801bb2e04ce061a6f873cb432bf1bb1e6072e295610904662ac427a30a",
-            transaction.getSignature().toString("hex"),
-        );
-        assert.equal(
-            transaction.getHash().toString(),
-            "32fb1681bd532b226b5bdeed61ae62ce9416bf5e92e48caf96253ff72d1670ac",
-        );
+        assert.throws(() => {
+            transaction.serializeForSigning();
+        }, "The `version` field should be 2 so the `options` field can be set.");
```
andreibancioiu, line 209: :+1:

andreibancioiu (commented):
A few optional comments, but it can also be :green_circle:

src/transactionComputer.ts
popenta: added a const `TRANSACTION_MIN_VERSION_THAT_SUPPORTS_OPTIONS`.

src/transactionComputer.ts
popenta: added.

src/transactionComputer.ts
popenta: Chenged!

src/transaction.spec.ts
popenta: Indeed, it worked well as it was before. Reverted the change and duplicated the test.

dragos-rebegea (approved):

PR #423, opened 2024-03-28
Implemented Address Computer

popenta:
Implemented the `AddressComputer` class according to the [sdk-specs](https://github.com/multiversx/mx-sdk-specs/blob/main/core/address.md#addresscomputer).

### DIFF
--- a/src/address.spec.ts
+++ b/src/address.spec.ts
@@ -1,5 +1,5 @@
 import { assert } from "chai";
-import { Address } from "./address";
+import { Address, AddressComputer } from "./address";
 import * as errors from "./errors";
 
 describe("test address", () => {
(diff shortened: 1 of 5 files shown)


src/smartcontracts/transactionPayloadBuilders.ts
```diff
@@ -1,11 +1,10 @@
+import { WasmVirtualMachine } from "../constants";
```
andreibancioiu, line 1: Should be fine, we can consider that this isn't a breaking change.

src/address.ts
```diff
@@ -259,3 +268,70 @@ export class Address {
...
 }
+
+export class AddressComputer {
+    private readonly numberOfShardsWithoutMeta: number;
+
+    constructor(numberOfShardsWithoutMeta?: number) {
+        this.numberOfShardsWithoutMeta = numberOfShardsWithoutMeta || CURRENT_NUMBER_OF_SHARDS_WITHOUT_META;
+    }
+
+    computeContractAddress(deployer: IAddress, deploymentNonce: bigint): Address {
```
andreibancioiu, line 279: Let's call this implementation from `smartContract.ts`.

src/address.ts
```diff
@@ -259,3 +268,70 @@ export class Address {
...
+        let initialPadding = Buffer.alloc(8, 0);
```
andreibancioiu, line 280: `const` where possible. Here, and below.

src/address.ts
```diff
@@ -259,3 +268,70 @@ export class Address {
...
+
+        let address = new Address(addressBytes);
+        return address;
+    }
+
+    getShardOfAddress(address: IAddress): number {
+        return this.getShardOfPubkey(address.getPublicKey(), this.numberOfShardsWithoutMeta);
+    }
+
+    private getShardOfPubkey(pubkey: Uint8Array, numberOfShards: number): number {
```
andreibancioiu, line 302: For review, linking to the Go and PY implementations:
 - https://github.com/multiversx/mx-chain-core-go/blob/main/core/sharding/multiShardCoordinator.go#L52
 - https://github.com/multiversx/mx-sdk-py-core/blob/main/multiversx_sdk_core/address.py#L141

This works for 1, 2, 3 and 4 shards. But will fail for more. See the Go implementation. For the moment, we can keep the PY and JS implementations as they are, but we should fix them in the future.

src/address.spec.ts
```diff
@@ -67,4 +67,33 @@ describe("test address", () => {
...
+        const addressComputer = new AddressComputer();
+        const deployer = Address.fromBech32("erd1j0hxzs7dcyxw08c4k2nv9tfcaxmqy8rj59meq505w92064x0h40qcxh3ap");
+
+        let contractAddress = addressComputer.computeContractAddress(deployer, 0n);
+        assert.equal(contractAddress.toHex(), "00000000000000000500bb652200ed1f994200ab6699462cab4b1af7b11ebd5e");
+        assert.equal(contractAddress.toBech32(), "erd1qqqqqqqqqqqqqpgqhdjjyq8dr7v5yq9tv6v5vt9tfvd00vg7h40q6779zn");
+
+        contractAddress = addressComputer.computeContractAddress(deployer, 1n);
+        assert.equal(contractAddress.toHex(), "000000000000000005006e4f90488e27342f9a46e1809452c85ee7186566bd5e");
+        assert.equal(contractAddress.toBech32(), "erd1qqqqqqqqqqqqqpgqde8eqjywyu6zlxjxuxqfg5kgtmn3setxh40qen8egy");
```
andreibancioiu, line 81: :+1: 

For reviewers: https://github.com/multiversx/mx-sdk-py-core/blob/main/multiversx_sdk_core/address_test.py

src/address.ts
popenta: Will replicate the Go implementation in the future.

src/smartcontracts/transactionPayloadBuilders.ts
popenta: I don't see why this could be breaking.

src/address.ts
popenta: Done!

src/address.ts
popenta: Done!

src/smartcontracts/smartContract.ts
```diff
@@ -259,26 +256,8 @@ export class SmartContract implements ISmartContract {
...
-        let addressBytes = Buffer.concat([
-            initialPadding,
-            vmTypeBytes,
-            hash.slice(10, 30),
-            shardSelector
-        ]);
-
-        let address = new Address(addressBytes);
-        return address;
+        const deployer = new Address(owner.bech32());
```
andreibancioiu, line 259: Could have used the new approach, `Address.fromBech32()`.

andreibancioiu (approved):

src/smartcontracts/smartContract.ts
popenta: indeed.

src/smartcontracts/transactionPayloadBuilders.ts
andreibancioiu: Being a publicly exported element, a movement can affect clients that use fully qualified imports (in this context, we can simply think of this as a non-breaking change, though).

andreibancioiu (approved):

danielailie (approved):

PR #432, opened 2024-04-05
add method to compute bytes for verifying transaction signature

popenta:


src/transaction.spec.ts
```diff
@@ -696,4 +697,60 @@ describe("test transaction", async () => {
         assert.equal(transaction.version, 2);
         assert.equal(transaction.options, 3);
     });
+
+    it("should compute bytes to verify transaction signature", async () => {
+        let transaction = new Transaction({
+            sender: wallets.alice.address.toBech32(),
```
andreibancioiu, line 703: For the sake of the examples, we can also add a transaction nonce - here and below.

andreibancioiu (approved):

src/transaction.spec.ts
popenta: added

andreibancioiu (approved):

bogdan-rosianu (approved):

PR #434, opened 2024-04-11
Added signer for the Message class

popenta:
Related to:
 - https://github.com/multiversx/mx-sdk-specs/pull/63

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -23,3 +23,4 @@ export const CURRENT_NUMBER_OF_SHARDS_WITHOUT_META = 3;
 export const WasmVirtualMachine = "0500";
 export const METACHAIN_ID = 4294967295;
 export const SDK_JS_SIGNER = "sdk-js";
+export const UNKNOWN_SIGNER = "unknown";
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -22,3 +22,4 @@ export const BECH32_ADDRESS_LENGTH = 62;
 export const CURRENT_NUMBER_OF_SHARDS_WITHOUT_META = 3;
 export const WasmVirtualMachine = "0500";
 export const METACHAIN_ID = 4294967295;
+export const SDK_JS_SIGNER = "sdk-js";
--- a/src/message.spec.ts
+++ b/src/message.spec.ts
@@ -1,6 +1,6 @@
 import { UserVerifier } from "@multiversx/sdk-wallet";
 import { assert } from "chai";
-import { DEFAULT_MESSAGE_VERSION, SDK_JS_SIGNER } from "./constants";
+import { DEFAULT_MESSAGE_VERSION, SDK_JS_SIGNER, UNKNOWN_SIGNER } from "./constants";
 import { Message, MessageComputer } from "./message";
 import { TestWallet, loadTestWallets } from "./testutils";
 
@@ -104,6 +104,6 @@ describe("test message", () => {
             "b16847437049986f936dd4a0917c869730cbf29e40a0c0821ca70db33f44758c3d41bcbea446dee70dea13d50942343bb78e74979dc434bbb2b901e0f4fd1809",
         );
         assert.deepEqual(message.version, DEFAULT_MESSAGE_VERSION);
-        assert.equal(message.signer, "unknown");
+        assert.equal(message.signer, UNKNOWN_SIGNER);
     });
 });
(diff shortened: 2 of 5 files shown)


src/constants.ts
```diff
@@ -22,3 +22,4 @@ export const BECH32_ADDRESS_LENGTH = 62;
 export const CURRENT_NUMBER_OF_SHARDS_WITHOUT_META = 3;
 export const WasmVirtualMachine = "0500";
 export const METACHAIN_ID = 4294967295;
+export const SDK_JS_SIGNER = "sdk-js";
```
andreibancioiu, line 25: I wonder whether we should also add the major version here. E.g. `sdk-js v13`.

src/message.ts
```diff
@@ -50,16 +61,24 @@ export class MessageComputer {
...
         };
     }
 
-    unpackMessage(packedMessage: { message: string; signature?: string; address?: string; version?: number }): Message {
+    unpackMessage(packedMessage: {
+        message: string;
+        signature?: string;
+        address?: string;
+        version?: number;
+        signer?: string;
```
andreibancioiu, line 80: Correct. Optional for now. Will be required in a future major release.

src/message.ts
```diff
@@ -72,12 +91,14 @@ export class MessageComputer {
         }
 
         const version = packedMessage.version || DEFAULT_MESSAGE_VERSION;
+        const signer = packedMessage.signer ?? SDK_JS_SIGNER;
```
andreibancioiu, line 94: Here, for unpack, actually, we should not use the default if missing. Maybe, instead, `unknown`?

src/message.ts
```diff
@@ -21,12 +21,23 @@ export class Message {
...
+        signature?: Uint8Array;
+        address?: IAddress;
+        version?: number;
+        signer?: string;
+    }) {
         this.data = options.data;
         this.signature = options.signature;
         this.address = options.address;
         this.version = options.version || DEFAULT_MESSAGE_VERSION;
+        this.signer = options.signer ?? SDK_JS_SIGNER;
```
andreibancioiu, line 40: On the constructor, here, it's OK to put the default is missing. But on unpack, we should use a different placeholder, I think. E.g. `unknown`.

src/message.ts
andreibancioiu: CC: @ccorcoveanu

popenta: That sounds good.

src/message.ts
popenta: set to `unknown` when unpacking a message with no `signer` field

popenta: indeed, here the `||` operator would have worked just fine, will replace it to stay consistent.  Above I am using the `||` operator because someone could have passed the value `0` which would be a valid number, but it's not a valid message version.

Bellow, this line `signer: message.signer ? message.signer : SDK_JS_SIGNER` could simply be `signer: message.signer` since `packMessage` needs an object of type `Message` and the signer always get's initialized in the constructor. The same applies for the `version`

src/message.ts
```diff
@@ -91,7 +91,7 @@ export class MessageComputer {
         }
 
         const version = packedMessage.version || DEFAULT_MESSAGE_VERSION;
-        const signer = packedMessage.signer ?? SDK_JS_SIGNER;
+        const signer = packedMessage.signer ?? "unknown";
```
andreibancioiu, line 94: Can be a constant, near the constant `SDK_JS_SIGNER`. E.g. `UNKNOWN_SIGNER`.

src/message.spec.ts
```diff
@@ -83,5 +85,25 @@ describe("test message", () => {
...
+
+        const message = messageComputer.unpackMessage(packedMessage);
+        assert.deepEqual(message.address, alice.getAddress());
+        assert.deepEqual(Buffer.from(message.data).toString(), "this is a test message");
+        assert.deepEqual(
+            Buffer.from(message.signature!).toString("hex"),
+            "b16847437049986f936dd4a0917c869730cbf29e40a0c0821ca70db33f44758c3d41bcbea446dee70dea13d50942343bb78e74979dc434bbb2b901e0f4fd1809",
+        );
+        assert.deepEqual(message.version, DEFAULT_MESSAGE_VERSION);
+        assert.equal(message.signer, "unknown");
```
andreibancioiu, line 107: :+1:

src/message.ts
popenta: added a const called `UNKNOWN_SIGNER`.

andreibancioiu (approved):

ccorcoveanu (approved):

PR #440, opened 2024-04-17
Changed ESDTTransferRole to be mandatory

popenta:
This in theory would be a **breaking** change but it's more of a **fixing** change.

### DIFF
--- a/src/transactionsFactories/tokenManagementTransactionsFactory.ts
+++ b/src/transactionsFactories/tokenManagementTransactionsFactory.ts
@@ -299,7 +299,7 @@ export class TokenManagementTransactionsFactory {
         tokenIdentifier: string;
         addRoleLocalMint: boolean;
         addRoleLocalBurn: boolean;
-        addRoleESDTTransferRole?: boolean;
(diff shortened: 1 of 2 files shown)


CiprianDraghici (approved):

radumojic (approved):

src/transactionsFactories/tokenManagementTransactionsFactory.ts
```diff
@@ -299,7 +299,7 @@ export class TokenManagementTransactionsFactory {
         tokenIdentifier: string;
         addRoleLocalMint: boolean;
         addRoleLocalBurn: boolean;
-        addRoleESDTTransferRole?: boolean;
+        addRoleESDTTransferRole: boolean;
```
andreibancioiu, line 302: Indeed. It's a change that slightly breaks the current public API, but **it's a fixing change**. Not passing the `addRoleESDTTransferRole` parameter is worse than being forced to (when upgrading to the new minor version of the library).

We should document this in the release notes, though.

andreibancioiu (approved):

PR #448, opened 2024-05-14
Align Address with sdk-specs, as much as possible (without breaking changes)

andreibancioiu:
- Align `Address` with the specs, as much as possible (but avoid breaking changes): https://github.com/multiversx/mx-sdk-specs/blob/main/core/address.md.
 - Small behavioral breaking change: adjust `ensureValidTransactionFields`, drop the logic that checks address length

### DIFF
--- a/src/address.ts
+++ b/src/address.ts
@@ -371,6 +371,7 @@ function decodeFromBech32(options: { value: string; allowCustomHrp: boolean }): 
     }
 
     // Workaround, in order to avoid behavioral breaking changes on legacy flows.
+    // In a future major release, we should drop this constraint (not exactly useful, validation should be performed in other ways)
     if (!allowCustomHrp && hrp != DEFAULT_HRP) {
         throw new errors.ErrAddressBadHrp(DEFAULT_HRP, hrp);
     }
--- a/src/address.ts
+++ b/src/address.ts
@@ -1,6 +1,7 @@
 import * as bech32 from "bech32";
 import BigNumber from "bignumber.js";
-import { CURRENT_NUMBER_OF_SHARDS_WITHOUT_META, DEFAULT_HRP, METACHAIN_ID, WasmVirtualMachine } from "./constants";
+import { LibraryConfig } from "./config";
+import { CURRENT_NUMBER_OF_SHARDS_WITHOUT_META, METACHAIN_ID, WasmVirtualMachine } from "./constants";
 import * as errors from "./errors";
 import { bigIntToBuffer } from "./tokenOperations/codec";
 const createKeccakHash = require("keccak");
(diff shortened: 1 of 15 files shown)


src/transactionsFactories/delegationTransactionsFactory.ts
```diff
@@ -8,6 +8,7 @@ import { TransactionBuilder } from "./transactionBuilder";
 
 interface IConfig {
     chainID: string;
+    addressHrp: string;
```
andreibancioiu, line 11: Negligible breaking change.

src/transactionsFactories/smartContractTransactionsFactory.ts
```diff
@@ -13,6 +13,7 @@ import { TransactionBuilder } from "./transactionBuilder";
 
 interface IConfig {
     chainID: string;
+    addressHrp: string;
```
andreibancioiu, line 16: Negligible breaking change.

src/transactionsFactories/tokenManagementTransactionsFactory.ts
```diff
@@ -8,6 +8,7 @@ import { TransactionBuilder } from "./transactionBuilder";
 
 interface IConfig {
     chainID: string;
+    addressHrp: string;
```
andreibancioiu, line 11: Negligible breaking change.

src/address.ts
```diff
@@ -140,7 +165,7 @@ export class Address {
         const prefix = decoded?.prefix;
         const pubkey = decoded ? Buffer.from(bech32.fromWords(decoded.words)) : undefined;
 
-        if (prefix !== HRP || pubkey?.length !== PUBKEY_LENGTH) {
+        if (prefix !== LibraryConfig.DefaultAddressHrp || pubkey?.length !== PUBKEY_LENGTH) {
```
popenta, line 168: why is the address considered not valid if the hrp is different than the default?

src/transactionsFactories/tokenManagementTransactionsFactory.ts
```diff
@@ -83,7 +86,7 @@ export class TokenManagementTransactionsFactory {
         return new TransactionBuilder({
             config: this.config,
             sender: options.sender,
-            receiver: Address.fromBech32(ESDT_CONTRACT_ADDRESS),
+            receiver: Address.fromHex(ESDT_CONTRACT_ADDRESS_HEX, this.config.addressHrp),
```
popenta, line 89: use `this.esdtContractAddress`

src/address.ts
andreibancioiu: Good question!

In order to avoid a breaking change with respect to the behavior of `isValid` (public method). We will change this in the future (a major version).

src/transactionsFactories/tokenManagementTransactionsFactory.ts
andreibancioiu: Fixed.

popenta (approved):

danielailie (approved):

danielailie (approved):

popenta (approved):

PR #462, opened 2024-06-28
Send Native token in MultiESDTNFTTransfer

popenta:
Added support to send both native and esdt tokens using `MultiESDTNFTTransfer`.

### DIFF
--- a/src/transactionsFactories/smartContractTransactionsFactory.ts
+++ b/src/transactionsFactories/smartContractTransactionsFactory.ts
@@ -96,9 +96,7 @@ export class SmartContractTransactionsFactory {
         let numberOfTokens = tokenTransfers.length;
 
         if (nativeTransferAmount && numberOfTokens) {
-            const nativeTransfer = TokenTransfer.newFromEgldAmount(nativeTransferAmount);
-            tokenTransfers.push(nativeTransfer);
-
+            tokenTransfers.push(TokenTransfer.newFromEgldAmount(nativeTransferAmount));
             nativeTransferAmount = 0n;
             numberOfTokens++;
         }
--- a/src/transactionsFactories/smartContractTransactionsFactory.ts
+++ b/src/transactionsFactories/smartContractTransactionsFactory.ts
@@ -91,7 +91,7 @@ export class SmartContractTransactionsFactory {
         tokenTransfers?: TokenTransfer[];
     }): Transaction {
         const args = options.arguments || [];
-        let tokenTransfers = options.tokenTransfers || [];
+        let tokenTransfers = options.tokenTransfers ? [...options.tokenTransfers] : [];
         let nativeTransferAmount = options.nativeTransferAmount ?? 0n;
         let numberOfTokens = tokenTransfers.length;
 
--- a/src/transactionsFactories/smartContractTransactionsFactory.ts
+++ b/src/transactionsFactories/smartContractTransactionsFactory.ts
@@ -1,12 +1,12 @@
 import { Address } from "../address";
-import { CONTRACT_DEPLOY_ADDRESS_HEX, VM_TYPE_WASM_VM } from "../constants";
-import { Err, ErrBadUsage } from "../errors";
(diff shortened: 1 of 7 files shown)


src/transactionsFactories/smartContractTransactionsFactory.ts
```diff
@@ -91,19 +91,23 @@ export class SmartContractTransactionsFactory {
...
-        const nativeTransferAmount = options.nativeTransferAmount ?? 0n;
-        const numberOfTokens = tokenTransfer.length;
+        let tokenTransfers = options.tokenTransfers ? [...options.tokenTransfers] : [];
+        let nativeTransferAmount = options.nativeTransferAmount ?? 0n;
+        let numberOfTokens = tokenTransfers.length;
 
         if (nativeTransferAmount && numberOfTokens) {
-            throw new ErrBadUsage("Can't send both native tokens and custom tokens(ESDT/NFT)");
+            const nativeTransfer = TokenTransfer.newFromEgldAmount(nativeTransferAmount);
+            tokenTransfers.push(nativeTransfer);
```
andreibancioiu, line 100: Maybe use array concatenation, similar to Python?

src/transactionsFactories/transferTransactionsFactory.ts
```diff
@@ -82,18 +82,10 @@ export class TransferTransactionsFactory {
         return this.gasEstimator !== undefined;
     }
 
-    private ensureMembersAreDefined() {
+    private ensureConfigIsDefined() {
```
andreibancioiu, line 85: Indeed, the presence of `this.config` indicates the presence of `tokenComputer` and `tokenTransfersDataBuilder`, as well. Also, we can set `this.tokenComputer` and `this.tokenTransfersDataBuilder` in constructor no matter what (irrespective to `isGasEstimator()`).

In v14, this is the first legacy thing to be dropped :)

src/transactionsFactories/transferTransactionsFactory.ts
```diff
@@ -152,6 +144,44 @@ export class TransferTransactionsFactory {
...
+            return this.createTransactionForNativeTokenTransfer({
+                sender: options.sender,
+                receiver: options.receiver,
+                nativeAmount: nativeAmount,
+                data: options.data,
+            });
+        }
+
+        const nativeTransfer = nativeAmount ? TokenTransfer.newFromEgldAmount(nativeAmount) : undefined;
+        nativeTransfer ? tokenTransfers.push(nativeTransfer) : null;
```
andreibancioiu, line 176: As mentioned above, array concatenation can be used (optional).

andreibancioiu (dismissed):

andreibancioiu (approved):

src/transactionsFactories/transferTransactionsFactory.ts
```diff
@@ -152,6 +144,46 @@ export class TransferTransactionsFactory {
...
+        nativeAmount?: bigint;
+        tokenTransfers?: TokenTransfer[];
+        data?: Uint8Array;
+    }): Transaction {
+        const nativeAmount = options.nativeAmount ?? 0n;
+        let tokenTransfers = options.tokenTransfers ? [...options.tokenTransfers] : [];
+        const numberOfTokens = tokenTransfers.length;
+
+        if (!nativeAmount && numberOfTokens === 0) {
+            throw new ErrBadUsage("No native token amount or token transfers provided");
```
bogdan-rosianu, line 159: will this be thrown if I want to send a 'notarizing' transaction? since the native amount is 0 and there is no transferred token. I believe that this is a valid use case and we should allow it. So only throw error if nativeAmount is undefined and tokenTransfers is undefined

bogdan-rosianu: ex: Alice wants to send Bob 0 egld but with data field 'Hello'. I think it will fail, as of now

popenta: It will, but added support for it now.

bogdan-rosianu (approved):

andreibancioiu (approved):

Issue #491, opened 2024-10-03
Improve support for parsing the outcome of contract calls within Relayed V3


andreibancioiu:
References:
 - https://t.me/MultiversXDevelopers/131109 (Telegram thread)

danielailie:
I'll close this as we have a different version for Relayed v3 and we don't need this anymore

PR #495, opened 2024-10-04
Allow one to parse the outcome of contract calls which are inner transactions (relayed V3)

andreibancioiu:
- See https://github.com/multiversx/mx-sdk-js-core/issues/491. 
- Implement `transactionOnNetworkToOutcomesOfInnerTransactions`.


src/converters/transactionsConverter.ts
```diff
@@ -141,4 +149,79 @@ export class TransactionsConverter {
             dataItems: dataItems,
         });
     }
+
+    /**
+     * Generally speaking, useful for Relayed V3 transactions.
+     */
+    public transactionOnNetworkToOutcomesOfInnerTransactions(
+        transactionOnNetwork: ITransactionOnNetwork,
```
andreibancioiu, line 157: Since the input is a `ITransactionOnNetwork`, this works both for the `TransactionOnNetwork` from the satellite package `sdk-network-providers`, and for the class within `sdk-core/networkProviders` (long time support rationale).

src/converters/transactionsConverter.ts
```diff
@@ -141,4 +149,79 @@ export class TransactionsConverter {
...
+        }
+
+        return outcomes;
+    }
+
+    /**
+     * Artificially converts an inner transaction (of a relayed V3) to a transaction on the network,
+     * by matching the inner transaction with its corresponding smart contract result.
+     */
+    private convertInnerTransactionToTransactionOnNetwork(
```
andreibancioiu, line 179: We'll keep this private, since it's a temporary workaround, until proper VM output is returned from the Proxy & API.

src/interfaceOfNetwork.ts
```diff
@@ -26,6 +26,10 @@ export interface ITransactionOnNetwork {
     receipt: ITransactionReceipt;
     contractResults: IContractResults;
     logs: ITransactionLogs;
+
+    // TODO: In a future major release, make these required (empty is allowed).
```
andreibancioiu, line 30: Actually, maybe, we'll completely drop these interfaces, and use simple DTOs (according to the specs).

src/networkProviders/interface.ts
```diff
@@ -143,22 +147,7 @@ export interface IAddress {
...
-    guardian: string;
-    signature: Uint8Array;
-    guardianSignature: Uint8Array;
-    relayer?: string;
-    innerTransactions?: ITransactionNext[];
-}
+/**
+ * @deprecated This will be remove with the next release (replaced by the `ITransaction` interface from "src/interface.ts").
+ */
+export type ITransactionNext = ITransactionAsInSpecs;
```
andreibancioiu, line 153: Previously, `ITransactionNext` was almost the same as `ITransactionAsInSpecs`, with the exception of:
 - `ITransactionNext.relayer` was optional, `ITransactionAsInSpecs.relayer` isn't (but empty is allowed)
 - `ITransactionNext.innerTransactions` was optional, `ITransactionAsInSpecs.innerTransactions` isn't (but empty is allowed).

This should not be considered a breaking change, since it's an alteration of an interface that's merely used as the substitute of `Transaction within the network provider components newly moved to `src/networkProviders`.

andreibancioiu:
Not needed anymore (relayed V3 will not stay in the current format).

PR #505, opened 2024-10-14
Make axios and bls optional

danielailie:


popenta (approved):

README.md
```diff
@@ -32,6 +32,24 @@ npm install --global browserify
 npm install esmify --no-save
 ```
 
+## Optional Dependencies
+
+### axios
+
+This package can make HTTP requests using `axios`, but it is not bundled by default. If you plan to use the API network provider or Proxy network provider, make sure to install `axios`:
```
andreibancioiu, line 39: `but it is not` -> `which is not` (optional) (here and below).

src/networkProviders/apiNetworkProvider.ts
```diff
@@ -202,21 +202,24 @@ export class ApiNetworkProvider implements INetworkProvider {
     }
 
     private async doGet(resourceUrl: string): Promise<any> {
+        let axios = await getAxios();
```
andreibancioiu, line 205: Can be `const` (here and below).

src/utils.ts
```diff
@@ -56,3 +56,13 @@ export function isEmpty(value: { isEmpty?: () => boolean; length?: number }): bo
 
     return value.length === 0;
 }
+
+export async function getAxios() {
+    let axios;
+    try {
+        axios = require("axios");
```
andreibancioiu, line 63: Can return directly, without declaring `let axios` above.

src/wallet/validatorKeys.ts
```diff
@@ -2,20 +2,29 @@ import { ErrInvariantFailed } from "../errors";
...
 import { parseValidatorKey } from "./pem";
 
-const bls = require("@multiversx/sdk-bls-wasm");
-
 export const VALIDATOR_SECRETKEY_LENGTH = 32;
 export const VALIDATOR_PUBKEY_LENGTH = 96;
 
 export class BLS {
     private static isInitialized: boolean = false;
+    public static bls: any;
```
andreibancioiu, line 10: A bit tricky. C#'s `internal` would have been nice here :)

Though, since we plan to drop Herumi's library, it's fine :+1: 

Alternative: private field and a public function `getUnderlyingComponent()` (or something similar).

src/networkProviders/apiNetworkProvider.ts
danielailie: done

src/utils.ts
danielailie: done

src/wallet/validatorKeys.ts
danielailie: I will let it like this since we plan to drop it

andreibancioiu (approved):

popenta (approved):

PR #514, opened 2024-10-17
Add entrypoints

danielailie:


src/transactionsFactories/accountTransactionsFactory.ts
```diff
@@ -3,7 +3,7 @@ import { IAddress } from "../interface";
 import { Transaction } from "../transaction";
 import { TransactionBuilder } from "./transactionBuilder";
 
-interface IConfig {
+export interface IConfig {
```
popenta, line 6: I don't think this should be exported. If we plan on using the same interfaces throughout the package (e.g. controllers) we should have an `interfaces.ts` file. If it's used in other places we should simply duplicate the inteface.

src/transactionsFactories/delegationTransactionsFactory.ts
```diff
@@ -376,6 +376,30 @@ export class DelegationTransactionsFactory {
         }).build();
     }
 
+    createTransactionForWithdrawing(_options: { sender: IAddress; delegationContract: IAddress }): Transaction {
```
popenta, line 379: perhaps add some empty lines, but it's not a big deal.

src/controllers/accountController.ts
```diff
@@ -0,0 +1,64 @@
+import { IAddress } from "../interface";
```
popenta, line 1: Since the methods `pubkey()` and `bech32()` are deprecated in the `Address` class, I think we can create locally a new interface that has `toBech32()` and `getPublicKey()`, given the fact that everybody will use that class.

src/controllers/smartContractController.ts
```diff
@@ -0,0 +1,129 @@
...
+    async createTransactionForUpgrade(sender: IAccount, options: ContractUpgradeInput): Promise<Transaction> {
+        const transaction = this.factory.createTransactionForUpgrade({ ...options, sender: sender.address });
+
+        transaction.nonce = options.nonce;
+        transaction.signature = await sender.sign(this.txComputer.computeBytesForSigning(transaction));
+
+        return transaction;
+    }
+
+    async createTransactionForExecute(sender: IAccount, options: TransactionInput): Promise<Transaction> {
```
popenta, line 61: I think you could've implemented `parseExecute` and `awaitExecuteCompleted`.

src/controllers/tokenManagementController.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    async awaitCompletedIssueFungible(txHash: string): Promise<IESDTIssueOutcome[]> {
+        const transaction = await this.transactionAwaiter.awaitCompleted(txHash);
+        return this.parseIssueFungible(transaction);
+    }
+
+    parseIssueFungible(transactionOnNetwork: TransactionOnNetwork): IESDTIssueOutcome[] {
+        return this.parser.parseIssueFungible(transactionOnNetwork);
+    }
```
popenta, line 43: add empty line

src/controllers/tokenManagementController.ts
popenta, line 1: Why didn't you implement all the methods that are present in `TokenManagementTransactionsFactory`?

src/controllers/interfaces.ts
```diff
@@ -0,0 +1,7 @@
+import { IAddress } from "../interface";
+
```
andreibancioiu, line 2: Indeed, here we hold interfaces to be used within controllers (we don't have the "package protected" modifier in TypeScript, unfortunately).

We've somehow inherited this pattern from Go codebases, but seems applicable in TypeScript as well - works fine for languages with structural typing (as opposed to languages nominal-typing only - C# / Java).

Nice stuff to read & debate (not necessarily off-topic): Go interfaces vs. Java interfaces; nominal typing vs. structural typing.

src/controllers/smartContractController.ts
```diff
@@ -0,0 +1,129 @@
...
+
+type ContractUpgradeInput = ContractDepoyInput & { contract: IAddress };
+
+interface SmartContractDeployOutcome {
+    returnCode: string;
+    returnMessage: string;
+    contracts: DeployedSmartContract[];
+}
+
+class DeployedSmartContract {
```
andreibancioiu, line 115: Tricky if it's defined as class. Maybe have it defined somewhere else, and used in the outcome parser itself (create instances and such)?

Tricky because: it's a class, and people might expect that `outcome.contracts` really are objects having `DeployedSmartContract` as prototype. But they won't have it (at runtime), they will be mere, plain objects. Maybe have it as type?

Thus, a few possible options:
1. have it as type instead of class
2. in the outcome parser, have returned objects be `DeployedSmartContract`
3. for deploy, re-map mere objects returned by the outcome parser to instances of your class
4. something else

src/entrypoints/entrypoints.ts
```diff
@@ -0,0 +1,144 @@
...
+import { TransactionWatcher } from "../transactionWatcher";
+import { UserVerifier } from "../wallet";
+import { DevnetEntrypointConfig, MainnetEntrypointConfig, TestnetEntrypointConfig } from "./config";
+import { ProviderWrapper } from "./providerWrapper";
+
+class NetworkEntrypoint {
+    private networkProvider: ApiNetworkProvider | ProxyNetworkProvider;
+    private chainId: string;
+
+    constructor(networkProviderUrl: string, networkProviderKind: string, chainId: string) {
```
andreibancioiu, line 24: Options pattern?

andreibancioiu, line 24: In TypeScript, we can have `networkProviderKind` as an enum or have the type annotation as `"proxy" | "api"` (can also stay as it is).

src/entrypoints/entrypoints.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    createAccountController(): AccountController {
+        return new AccountController(this.chainId);
+    }
+
+    createRelayedController(): RelayedController {
+        return new RelayedController(this.chainId);
+    }
+
+    createSmartContractController(abi?: AbiRegistry): SmartContractController {
```
andreibancioiu, line 100: We can have an alias `Abi`, for example:

```
class Abi extends AbiRegistry;
```

src/controllers/tokenManagementController.ts
```diff
@@ -0,0 +1,108 @@
...
+import { INetworkProvider } from "../networkProviders/interface";
+import { IESDTIssueOutcome } from "../tokenOperations";
+import { Transaction } from "../transaction";
+import { TransactionComputer } from "../transactionComputer";
+import { TokenManagementTransactionsFactory, TransactionsFactoryConfig } from "../transactionsFactories";
+import { TokenManagementTransactionsOutcomeParser } from "../transactionsOutcomeParsers";
+import { TransactionWatcher } from "../transactionWatcher";
+import { IAccount } from "./interfaces";
+
+export class TokenManagementController {
```
andreibancioiu, line 12: Seems incomplete. But we can add remaining functionality in future PRs.

src/entrypoints/entrypoints.spec.ts
```diff
@@ -0,0 +1,105 @@
+import { assert } from "chai";
+import { readFileSync } from "fs";
+import { Account } from "../account";
+import { Address } from "../address";
+import { loadAbiRegistry, loadTestWallet, TestWallet } from "../testutils";
+import { TransactionComputer } from "../transactionComputer";
+import { DevnetEntrypoint } from "./entrypoints";
+
+describe("TestEntrypoint", () => {
+    const entrypoint = new DevnetEntrypoint();
```
andreibancioiu, line 10: :rocket:

andreibancioiu (commented):
:rocket: :moon:

src/controllers/delegationController.ts
```diff
@@ -0,0 +1,280 @@
...
+        const transaction = this.factory.createTransactionForWithdrawing({ ...options, sender: sender.address });
+
+        transaction.nonce = options.nonce;
+        transaction.signature = await sender.sign(this.txComputer.computeBytesForSigning(transaction));
+
+        return transaction;
+    }
+}
+
+type NewDelegationContractInput = { nonce: bigint; totalDelegationCap: bigint; serviceFee: bigint; amount: bigint };
```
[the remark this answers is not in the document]
danielailie: I don't think we will export them

src/address.ts
```diff
@@ -298,7 +294,7 @@ export class AddressComputer {
 
     computeContractAddress(deployer: IAddress, deploymentNonce: bigint): Address {
         const initialPadding = Buffer.alloc(8, 0);
-        const ownerPubkey = deployer.getPublicKey();
+        const ownerPubkey = deployer.pubkey();
```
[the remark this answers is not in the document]
danielailie: IAddress has only pubkey, we discussed that we will drop that one so.. no need to add extra implementation

danielailie:
this was done in smaller prs

PR #526, opened 2024-11-07
Add prefix handling for esdt prefix

danielailie:


src/tokens.ts
```diff
@@ -249,19 +249,31 @@ export class TokenComputer {
 
     extractIdentifierFromExtendedIdentifier(identifier: string): string {
         const parts = identifier.split("-");
+        const { prefix, ticker, randomnes } = this.splitIdentifierIntoComponent(parts);
```
popenta, line 252: typo in `randomness`. can also be renamed to `randomSequence`.

src/tokens.ts
```diff
@@ -281,6 +293,15 @@ export class TokenComputer {
         }
     }
 
+    private checkLengthOfPrefix(prefix: string): void {
+        const TOKEN_PREFIX_LENGTH = 4;
+        if (prefix.length > TOKEN_PREFIX_LENGTH) {
```
popenta, line 298: maybe we should also check that the length of the prefix is at least `1`.

popenta (approved):

src/tokens.ts
```diff
@@ -249,19 +249,31 @@ export class TokenComputer {
...
+        this.checkLengthOfRandomSequence(randomnes);
+        if (prefix) {
+            this.checkLengthOfPrefix(prefix);
+            return prefix + "-" + ticker + "-" + randomnes;
+        }
+        return ticker + "-" + randomnes;
+    }
+
+    splitIdentifierIntoComponent(parts: string[]): { prefix: any; ticker: any; randomnes: any } {
+        if (parts.length >= 3 && parts[2].length === 6) {
```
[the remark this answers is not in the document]
andreibancioiu: How would it be to contain the `const parts = identifier.split("-");` itself within this function, too? :cloud_with_snow:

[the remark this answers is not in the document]
danielailie: I don't know what name should I use for 3, any suggestion?

src/tokens.ts
```diff
@@ -249,19 +250,32 @@ export class TokenComputer {
...
+        if (prefix) {
+            this.checkLengthOfPrefix(prefix);
+            return prefix + "-" + ticker + "-" + randomSequence;
+        }
+        return ticker + "-" + randomSequence;
+    }
+
+    private splitIdentifierIntoComponents(identifier: string): { prefix: any; ticker: any; randomSequence: any } {
+        const parts = identifier.split("-");
+        if (parts.length >= 3 && parts[2].length === this.TOKEN_RANDOM_SEQUENCE_LENGTH) {
```
axenteoctavian, line 267: This condition is not fully correct. `NFT-1q2w3e-100000` will be considered with prefix and is not correct.
Another way maybe is to check:
- if parts[0] is full lowercase alphanumeric => starts with prefix
- else if parts[0] is full uppercase alphanumeric => starts with ticker
- otherwise throw error

src/tokens.ts
```diff
@@ -235,33 +236,50 @@ export class TokenComputer {
...
-        if (parts.length == 2) {
+        // If identifier is for a fungible token (2 parts or 3 with prefix), return 0
+        if (parts.length === 2 || (prefix && parts.length === 3)) {
             return 0;
         }
 
-        const hexNonce = Buffer.from(parts[2], "hex");
-        return decodeUnsignedNumber(hexNonce);
+        // Otherwise, decode the last part as an unsigned number
+        const hecNonce = parts[parts.length - 1];
```
andreibancioiu: Typo.

src/tokens.ts
```diff
@@ -235,33 +236,50 @@ export class TokenComputer {
...
     }
 
     extractIdentifierFromExtendedIdentifier(identifier: string): string {
         const parts = identifier.split("-");
+        const { prefix, ticker, randomSequence } = this.splitIdentifierIntoComponents(identifier);
+
+        this.validateExtendedIdentifier(prefix, randomSequence, ticker, parts);
+        if (prefix) {
+            this.checkLengthOfPrefix(prefix);
+            return prefix + "-" + ticker + "-" + randomSequence;
```
andreibancioiu, line 259: We could have had string interpolation / optional :upside_down_face:

src/tokens.ts
```diff
@@ -236,46 +236,50 @@ export class TokenComputer {
...
-        if (parts.length == 2) {
+        // If identifier is for a fungible token (2 parts or 3 with prefix), return 0
+        if (parts.length === 2 || (prefix && parts.length === 3)) {
             return 0;
         }
 
-        const hexNonce = Buffer.from(parts[2], "hex");
-        return decodeUnsignedNumber(hexNonce);
+        // Otherwise, decode the last part as an unsigned number
+        const hecNonce = parts[parts.length - 1];
```
popenta, line 248: typo

src/tokens.ts
```diff
@@ -236,46 +236,50 @@ export class TokenComputer {
...
-        this.checkLengthOfRandomSequence(randomSequence);
+        this.validateExtendedIdentifier(prefix, randomSequence, ticker, parts);
         if (prefix) {
             this.checkLengthOfPrefix(prefix);
             return prefix + "-" + ticker + "-" + randomSequence;
         }
         return ticker + "-" + randomSequence;
     }
 
+    private validateExtendedIdentifier(prefix: string, randomSequence: string, ticker: string, parts: string[]): void {
```
popenta, line 264: prefix can be `string | null`

popenta: also reorder args to represent the token structure (prefix, ticker, random sequence)

src/tokens.ts
```diff
@@ -235,33 +236,50 @@ export class TokenComputer {
...
-        if (parts.length == 2) {
+        // If identifier is for a fungible token (2 parts or 3 with prefix), return 0
+        if (parts.length === 2 || (prefix && parts.length === 3)) {
             return 0;
         }
 
-        const hexNonce = Buffer.from(parts[2], "hex");
-        return decodeUnsignedNumber(hexNonce);
+        // Otherwise, decode the last part as an unsigned number
+        const hecNonce = parts[parts.length - 1];
```
axenteoctavian, line 248:
```suggestion
        const strNonce = parts[parts.length - 1];
```

popenta: `hexNonce` would be a better name, in my opinion.

popenta (approved):

axenteoctavian (approved):

andreibancioiu (approved):

PR #527, opened 2024-11-09
Temporary workaround (hotfix): do not call "extendUserAgent"

andreibancioiu:
Temporary workaround (hotfix): do not call `extendUserAgent`, until a complete solution is found.


andreibancioiu:
Fixed separately: https://github.com/multiversx/mx-sdk-js-core/pull/528.

PR #541, opened 2024-12-09
Remove deprecated code

danielailie:


src/accounts/account.ts
```diff
@@ -18,14 +18,6 @@ export class Account implements IAccount {
      */
     nonce: INonce = 0;
```
popenta, line 19: this should also be changed to `BigInt`.

popenta: And at some point, the class should be updated to specs.

src/networkProviders/apiNetworkProvider.ts
```diff
@@ -146,17 +147,17 @@ export class ApiNetworkProvider implements INetworkProvider {
         return status;
     }
 
-    async sendTransaction(tx: ITransaction | ITransactionNext): Promise<string> {
+    async sendTransaction(tx: ITransaction): Promise<string> {
```
popenta, line 150: I think `ITransactionNext` should have been deleted only when updating the `Transaction` class and use that here.

popenta: I guess it's fine since you'll do it in the next PR.

popenta (approved):

src/accounts/account.ts
danielailie: if it's ok I'll do it in a different pr to keep this for remove only

andreibancioiu (approved):

PR #542, opened 2024-12-10
Remove ITransaction interface and update Transaction to follow specs

danielailie:


src/transaction.local.net.spec.ts
```diff
@@ -40,16 +39,16 @@ describe("test transaction", function () {
         let transactionOne = new Transaction({
             sender: alice.address,
             receiver: bob.address,
-            value: TokenTransfer.egldFromAmount(42),
-            gasLimit: network.MinGasLimit,
+            value: TokenTransfer.egldFromAmount(42).amount,
```
popenta, line 42: let's not use this method

danielailie: I'll update in a different pr if that's ok

src/proto/serializer.spec.ts
```diff
@@ -59,12 +57,12 @@ describe("serialize transactions", () => {
 
     it("with data, with value", async () => {
         let transaction = new Transaction({
-            nonce: 91,
-            value: TokenTransfer.egldFromAmount(10),
+            nonce: 91n,
+            value: TokenTransfer.egldFromAmount(10).amount,
```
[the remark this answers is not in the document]
danielailie: I'll update in a different pr if that's ok

src/abi/smartContract.ts
```diff
@@ -289,7 +289,7 @@ export class SmartContract implements ISmartContract {
      * @param nonce The owner nonce used for the deployment transaction
      */
     static computeAddress(owner: Address, nonce: INonce): Address {
-        const deployer = Address.fromBech32(owner.bech32());
+        const deployer = Address.fromBech32(owner.toBech32());
```
popenta, line 292: here it seems that we create an Address from an Address.

popenta (approved):

src/abi/nativeSerializer.ts
```diff
@@ -403,9 +403,9 @@ export namespace NativeSerializer {
     export function convertNativeToAddress(
         native: NativeTypes.NativeAddress,
         errorContext: ArgumentErrorContext,
-    ): IAddress {
+    ): Address {
         if ((<any>native).bech32) {
-            return <IAddress>native;
+            return <Address>native;
```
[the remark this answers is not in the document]
danielailie: after internal discussion it is ok as it is

andreibancioiu (approved):

PR #543, opened 2024-12-12
Tool 393 add relayer v 3 support

danielailie:


src/converters/transactionsConverter.ts
```diff
@@ -46,6 +49,7 @@ export class TransactionsConverter {
             nonce: BigInt(object.nonce),
             value: BigInt(object.value || ""),
             receiver: object.receiver,
+            relayer: object.relayer ? Address.newFromBech32(object.relayer) : Address.empty(),
```
andreibancioiu, line 52: Can we have it the same as `guardian`? If so, which way?

src/proto/serializer.ts
```diff
@@ -60,6 +60,11 @@ export class ProtoSerializer {
             protoTransaction.GuardianSignature = transaction.guardianSignature;
         }
 
+        if (transaction.relayer && !transaction.relayer.isEmpty()) {
```
andreibancioiu, line 63: Can refactor to have a mini private function `isRelayedTransaction` (for symmetry) .

src/proto/serializer.ts
```diff
@@ -60,6 +60,11 @@ export class ProtoSerializer {
...
+            protoTransaction.Relayer = transaction.relayer.getPublicKey();
+            protoTransaction.RelayerSignature = transaction.relayerSignature;
+        }
```
popenta, line 66: Let's extract to a function similar to the guardian check above, and also check if relayer signature is set.

src/transaction.spec.ts
```diff
@@ -751,4 +751,44 @@ describe("test transaction", async () => {
         assert.equal(isSignedByAlice, true);
         assert.equal(isSignedByBob, false);
     });
+
+    it("should serialize transaction with relayer", async () => {
+        const transaction = new Transaction({
+            chainID: networkConfig.ChainID,
+            sender: wallets.alice.address.bech32(),
```
popenta, line 758: use `toBech32()`, here and bellow.

src/transaction.ts
```diff
@@ -92,6 +91,11 @@ export class Transaction {
      */
     public guardian: string;
 
+    /**
+     * The relayer, in address format, next version all the other addresses will not be string anymore.
```
andreibancioiu, line 95: Can be rephrased.

```
The relayer address.
Note: in the next major version, `sender`, `receiver` and `guardian` will also have the type `Address`, instead of `string`.
```

src/transaction.ts
```diff
@@ -137,9 +148,11 @@ export class Transaction {
         this.version = Number(options.version?.valueOf() || TRANSACTION_VERSION_DEFAULT);
         this.options = Number(options.options?.valueOf() || TRANSACTION_OPTIONS_DEFAULT);
         this.guardian = options.guardian ? this.addressAsBech32(options.guardian) : "";
+        this.relayer = options.relayer?.isEmpty() ? undefined : options.relayer;
```
andreibancioiu, line 151: Oh, now I see, cannot be same as guardian, since that's a string :+1:

src/converters/transactionsConverter.ts
```diff
@@ -25,9 +26,11 @@ export class TransactionsConverter {
             chainID: transaction.chainID.valueOf(),
             version: transaction.version,
             options: transaction.options == 0 ? undefined : transaction.options,
+            relayer: !transaction.relayer || transaction.relayer.isEmpty() ? undefined : transaction.relayer.toBech32(),
```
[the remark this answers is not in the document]
andreibancioiu: Oh, now I see, cannot be same as guardian, since that's a string :+1:

src/converters/transactionsConverter.ts
andreibancioiu: Oh, now I see, cannot be same as guardian, since that's a string :+1:

src/transaction.ts
```diff
@@ -92,6 +91,11 @@ export class Transaction {
      */
     public guardian: string;
 
+    /**
+     * The relayer, in address format, next version all the other addresses will not be string anymore.
+     */
+    public relayer?: Address;
```
popenta, line 97: maybe this doesn't need to be optional. Make it optional on the constructor and if not provided, set to `Address.empty()`?

src/transactionComputer.ts
```diff
@@ -94,6 +95,17 @@ export class TransactionComputer {
         transaction.guardian = guardian;
     }
 
+    applyRelayer(transaction: ITransaction, relayer: Address) {
```
andreibancioiu, line 98: In sdk-py, we don't have it:

https://github.com/multiversx/mx-sdk-py/blob/feat/next/multiversx_sdk/core/transaction_computer.py#L83

Should we have it?

https://github.com/multiversx/mx-sdk-specs/blob/main/core/transaction_computer.md

src/transactionComputer.ts
```diff
@@ -94,6 +95,17 @@ export class TransactionComputer {
...
+        transaction.relayer = relayer;
+    }
```
popenta, line 100: this is kind of useless since one can simply do:
```js
transaction.relayer = relayerAddress;
```

src/transactionComputer.ts
```diff
@@ -94,6 +95,17 @@ export class TransactionComputer {
...
+
+    isRelayedV3Transaction(transaction: ITransaction) {
+        if (transaction.relayer && !transaction.relayer.isEmpty()) {
```
andreibancioiu, line 103: Can use a one-line return (can also stay as it is).

src/transaction.ts
andreibancioiu, line 97: Indeed, optional - I see we have the same in PY :+1: 

https://github.com/multiversx/mx-sdk-py/blob/feat/next/multiversx_sdk/core/transaction.py#L49

src/proto/serializer.ts
danielailie: we discussed about this internally and said it is not needed

src/transactionComputer.ts
popenta: I suggest we delete it from specs, as well. It is useless.

popenta (approved):

andreibancioiu (approved):

MWFIAE:
@danielailie @andreibancioiu 

What about the `RelayedTransactionsFactory`? It does not contain a `createRelayedV3Transaction` method, that was included in previous versions.

PR #551, opened 2024-12-17
Update token transfer to follow specs

danielailie:


andreibancioiu (approved):

andreibancioiu (approved):

src/tokens.ts
```diff
@@ -85,7 +85,7 @@ export class TokenTransfer {
         }
     }
 
-    static newFromEgldAmount(amount: bigint): TokenTransfer {
```
popenta, line 88: maybe this should have been deprecated as well? So it is not breaking.

popenta (approved):

src/tokens.ts
danielailie: we discussed in the office and said it is ok to rename it as it was not long time added

PR #559, opened 2025-01-21
Add account awaiter

danielailie:


src/networkProviders/accountAwaiter..dev.net.spec.ts
popenta, line 1: filename contains 2 consecutive dots.

src/networkProviders/accountAwaiter.ts
```diff
@@ -0,0 +1,104 @@
...
+            try {
+                fetchedData = await doFetch();
+                isConditionSatisfied = isSatisfied(fetchedData);
+
+                if (isConditionSatisfied) {
+                    break;
+                }
+            } catch (ex) {
+                throw ex;
+            }
```
andreibancioiu, line 83: Is this what we want?

danielailie: this is what we are doing on the transaction watcher also..

andreibancioiu: All right then for now. We should check if that's the best approach - for both of them (separately, another task / PR).

andreibancioiu (approved):

popenta (approved):

PR #573, opened 2025-02-13
Update token ticker validation rules

danielailie:


package.json
```diff
@@ -81,4 +81,4 @@
     "@multiversx/sdk-bls-wasm": "0.3.5",
     "bip39": "3.1.0"
   }
-}
+}
```
popenta, line 84: add empty line

src/tokens.spec.ts
```diff
@@ -46,6 +46,22 @@ describe("test tokens and token computer", async () => {
         const fungibleTokenIdentifier = "FNG-123456";
         identifier = tokenComputer.extractIdentifierFromExtendedIdentifier(fungibleTokenIdentifier);
         assert.equal(identifier, "FNG-123456");
+
+        const numericTokenTicker = "2024-45a190";
```
popenta, line 50: can we double check if lowercase ticker is accepted on the network?

andreibancioiu, line 50: Maybe use a more artificial example (not the token from mainnet).

andreibancioiu: Code is here:
 - https://github.com/multiversx/mx-chain-go/blob/master/vm/systemSmartContracts/esdt.go#L721

popenta: Indeed, thank you.

andreibancioiu (approved):

popenta (approved):

PR #574, opened 2025-02-24
Add relayer and guardian to controllers

danielailie:


andreibancioiu (approved):
I wonder if we should have them as regular optional arguments, _missing_ by default, instead of "empty" addresses.

src/abi/interaction.spec.ts
```diff
@@ -204,8 +204,7 @@ describe("test smart contract interactor", function () {
     });
 
     it("should interact with 'answer'", async function () {
-        setupUnitTestWatcherTimeouts();
-
+        this.timeout(30000);
```
andreibancioiu, line 207: This file contains unit tests - we can skip setting the timeout.

andreibancioiu (dismissed):

src/accountManagement/accountController.ts
```diff
@@ -32,28 +35,41 @@ export class AccountController {
...
         transaction.signature = await sender.sign(this.txComputer.computeBytesForSigning(transaction));
 
         return transaction;
     }
 
-    async createTransactionForGuardingAccount(sender: IAccount, nonce: bigint): Promise<Transaction> {
+    async createTransactionForGuardingAccount(
+        sender: IAccount,
+        nonce: bigint,
+        options: { guardian?: Address; relayer?: Address },
```
popenta, line 53: here, the guardian in not needed, should be removed.

popenta (changes requested):

popenta (approved):

PR #595, opened 2025-04-14
Add multisig transaction factory

danielailie:


src/multisig/multisigTransactionsFactory.ts
```diff
@@ -0,0 +1,509 @@
...
+import { Transaction } from "../core/transaction";
+import { TransactionBuilder } from "../core/transactionBuilder";
+import { SmartContractTransactionsFactory } from "../smartContracts";
+import * as resources from "./resources";
+
+interface IAbi {
+    constructorDefinition: EndpointDefinition;
+    upgradeConstructorDefinition?: EndpointDefinition;
+
+    getEndpoint(name: string | ContractFunction): EndpointDefinition;
```
popenta, line 28: I think type of `name` should only be string. As it is in the current implementation.

src/multisig/multisigTransactionsFactory.ts
```diff
@@ -0,0 +1,509 @@
...
+            this.argSerializer.valuesToStrings([new AddressValue(options.boardMemberAddress)])[0],
+        ];
+
+        return new TransactionBuilder({
+            config: this.config,
+            sender: sender,
+            receiver: options.multisigContract,
+            dataParts: dataParts,
+            gasLimit: options.gasLimit,
+            addDataMovementGas: true,
```
popenta, line 80: I think we should not add extra gas limit. Since the gas is provided by the user, he should consider everything. Same thing for the methods bellow.

src/multisig/multisigTransactionsFactory.ts
```diff
@@ -0,0 +1,509 @@
...
+            tokenTransfers: [],
+            functionName: options.functionName,
+            arguments: options.functionArguments,
+            abi: options.abi,
+        });
+        const dataParts = [
+            "proposeTransferExecute",
+            this.argSerializer.valuesToStrings([new AddressValue(options.to)])[0],
+            this.argSerializer.valuesToStrings([new BigUIntValue(options.egldAmount)])[0],
+            this.argSerializer.valuesToStrings([new OptionValue(new OptionType(new U64Type()), gasOption)])[0],
```
popenta, line 161: can you perhaps call `valuesToStrings()` only once and pass all the arguments at once?

src/multisig/multisigTransactionsFactory.ts
```diff
@@ -0,0 +1,509 @@
...
+                amount: new BigUIntValue(token.amount),
+            });
+        }
+        const dataParts = [
+            "proposeTransferExecuteEsdt",
+            ...this.argSerializer.valuesToStrings(
+                NativeSerializer.nativeToTypedValues(
+                    [options.to, argsTyped, options.gasLimit, VariadicValue.fromItems(...input.functionCall)],
+                    this.abi?.getEndpoint("proposeTransferExecuteEsdt") ??
+                        new EndpointDefinition("proposeTransferExecuteEsdt", [], [], new EndpointModifiers("", [])),
```
popenta, line 221: Does this work if the Abi is not provided? 
Perhaps you could convert all values to typed values so the abi would not be needed.

src/multisig/multisigTransactionsFactory.ts
```diff
@@ -0,0 +1,509 @@
...
+            functionName: options.functionName,
+            arguments: options.functionArguments,
+            abi: options.abi,
+        });
+        let receiver = options.multisigContract;
+        const dataParts = [
+            "proposeAsyncCall",
+            this.argSerializer.valuesToStrings([new AddressValue(options.to)])[0],
+            this.argSerializer.valuesToStrings([new BigUIntValue(options.nativeTransferAmount)])[0],
+            this.argSerializer.valuesToStrings([new BigUIntValue(options.gasLimit ?? 0n)])[0],
```
popenta, line 252: same thing as above, to call `valuesToStrings` once, if possible.

src/multisig/multisigTransactionsFactory.spec.ts
```diff
@@ -0,0 +1,529 @@
...
+        });
+    });
+
+    it("should create transaction for deploy multisig contract", function () {
+        const senderAddress = Address.newFromBech32("erd1spyavw0956vq68xj8y4tenjpq2wd5a9p2c6j8gsz7ztyrnpxrruqzu66jx");
+
+        const boardMemberAddress = Address.newFromBech32(
+            "erd1spyavw0956vq68xj8y4tenjpq2wd5a9p2c6j8gsz7ztyrnpxrruqzu66jx",
+        );
+        const proposerAddress = Address.newFromBech32("erd1k2s324ww2g0yj38qn2ch2jwctdy8mnfxep94q9arncc6xecg3xaq6mjse8");
```
popenta, line 36: this is a board member, as well.

popenta (approved):

src/multisig/multisigTransactionsFactory.ts
andreibancioiu: Maybe we can already deprecate `ContractFunction` in v14. And perhaps we'll remove it in v15.

src/multisig/proposeTransferExecuteContract.ts
```diff
@@ -0,0 +1,90 @@
...
+            arguments: options.arguments,
+            nativeTransferAmount: 0n,
+        });
+        const functionCallParts = Buffer.from(transaction.data).toString().split(ARGUMENTS_SEPARATOR);
+        const functionName = functionCallParts[0];
+        const functionArguments = [];
+        for (let index = 1; index < functionCallParts.length; index++) {
+            const element = functionCallParts[index];
+            functionArguments.push(element.valueOf());
+        }
```
andreibancioiu, line 81: Does `Array.slice` help here?

src/multisig/proposeTransferExecuteContract.ts
```diff
@@ -0,0 +1,90 @@
...
+            abi: options.abi,
+        });
+        const transaction = transactionsFactory.createTransactionForExecute(Address.empty(), {
+            contract: Address.empty(),
+            function: options.functionName,
+            gasLimit: 0n,
+            arguments: options.arguments,
+            nativeTransferAmount: 0n,
+        });
+        const functionCallParts = Buffer.from(transaction.data).toString().split(ARGUMENTS_SEPARATOR);
```
andreibancioiu, line 75: Maybe some newlines to separate the code blocks - here and above.

src/multisig/proposeTransferExecuteContract.ts
```diff
@@ -0,0 +1,90 @@
...
+        const functionName = functionCallParts[0];
+        const functionArguments = [];
+        for (let index = 1; index < functionCallParts.length; index++) {
+            const element = functionCallParts[index];
+            functionArguments.push(element.valueOf());
+        }
```
andreibancioiu, line 45: As below, in case `Array.slice` can be used.

src/multisig/multisigTransactionsFactory.ts
```diff
@@ -0,0 +1,510 @@
...
+        const input = ProposeTransferExecuteContractInput.newFromTransferExecuteInput({
+            multisig: options.multisigContract,
+            to: options.to,
+            functionName: options.functionName,
+            arguments: options.functionArguments,
+            abi: options.abi,
+        });
+        const tokenComputer = new TokenComputer();
+        const argsTyped = [];
+        for (const token of options.tokens) {
```
andreibancioiu, line 204: Maybe some newlines for better separation of code blocks / here / below.

src/multisig/multisigTransactionsFactory.spec.ts
```diff
@@ -0,0 +1,523 @@
+import { assert } from "chai";
+import { Abi, Code } from "../abi";
+import { CodeMetadata, Token, TokenTransfer } from "../core";
+import { Address } from "../core/address";
+import { Transaction } from "../core/transaction";
+import { TransactionsFactoryConfig } from "../core/transactionsFactoryConfig";
+import { loadAbiRegistry, loadContractCode } from "../testutils";
+import { MultisigTransactionsFactory } from "./multisigTransactionsFactory";
+
+describe("test multisig transactions factory", function () {
```
andreibancioiu, line 10: :rocket:

andreibancioiu (commented):
:rocket:

src/multisig/multisigTransactionsFactory.ts
```diff
@@ -0,0 +1,510 @@
...
+    constructorDefinition: EndpointDefinition;
+    upgradeConstructorDefinition?: EndpointDefinition;
+
+    getEndpoint(name: string): EndpointDefinition;
+}
+/**
+ * Use this class to create multisig related transactions like creating a new multisig contract,
+ * proposing actions, signing actions, and performing actions.
+ */
+export class MultisigTransactionsFactory extends SmartContractTransactionsFactory {
```
andreibancioiu, line 34: Inheriting vs. doing "composition" - debatable; but here, indeed, using `SmartContractTransactionsFactory` as a base class seems the better choice :+1:

src/multisig/multisigTransactionsFactory.ts
danielailie: not, but I'll do this in a different pr

src/multisig/multisigTransactionsFactory.ts
```diff
@@ -230,6 +216,21 @@ export class MultisigTransactionsFactory extends SmartContractTransactionsFactor
         }).build();
     }
 
+    private mapTokenPayment(options: resources.ProposeTransferExecuteEsdtInput) {
```
andreibancioiu, line 219: Can be plural, since there are more payments.

Also, we should specify the return type.

src/multisig/multisigTransactionsFactory.ts
```diff
@@ -230,6 +216,21 @@ export class MultisigTransactionsFactory extends SmartContractTransactionsFactor
...
+        const tokenComputer = new TokenComputer();
+        const argsTyped = [];
+        for (const token of options.tokens) {
+            argsTyped.push({
+                token_identifier: new TokenIdentifierValue(
```
andreibancioiu, line 224: When ABI is available, I think the following is sufficient:

```
{
    token_identifier: tokenComputer.extractIdentifierFromExtendedIdentifier(token.token.identifier),
    token_nonce: token.token.nonce,
    amount: token.amount
}
```

When the ABI isn't available, this does not work at all (as far as I remember), since a `StructValue` is required - a mere `{}` object does not suffice.

These being said, `argsTyped` can also be renamed (here, the args aren't completely typed).

Additional comments about having the ABI as required or not can be seen in the PR with the controller implementation.

src/multisig/proposeTransferExecuteContractInput.ts
```diff
@@ -72,19 +67,22 @@ export class ProposeTransferExecuteContractInput {
...
+
         return new ProposeTransferExecuteContractInput({
             multisigContract: options.multisig,
             to: options.to,
             functionCall: functionCall,
             gasLimit: options.optGasLimit,
         });
     }
+
+    private static getFunctionCall(transaction: Transaction) {
```
andreibancioiu, line 81: We should add a comment to explain the workaround / hack.

> In the multisig transactions factory, we leverage the generic smart contract transactions factory to properly encode the arguments (given the ABI) for the "other" contract execute proposal. However, since we have to handle them as `variadic<bytes>` when preparing the "top-level" multisig transaction, here we split back the data.

To-do for a future PR: convert the arguments to a list of buffers using the `ArgSerializer` or a `SmartContractTransactionsFactory.argsToDataParts()`, instead of the _<create transaction using factory, then parse back into list of buffers>_ hack.

popenta: As discussed, this is ok for the moment as long as we add the comment. Will perhaps refactor later if needed.

andreibancioiu (approved):

popenta (approved):

popenta (dismissed):

popenta (approved):

andreibancioiu (approved):

PR #602, opened 2025-05-04
Fix contract results parsing

michavie:
Test incorrectly assumes that contract results are returned in hex – it's base64 instead.

Quick debugging insight showcasing the invalid base64 <> hex comparison:

Screenshot 2025-05-04 at 11 44 05 PM

### DIFF
--- a/src/smartContracts/smartContractTransactionsOutcomeParser.spec.ts
+++ b/src/smartContracts/smartContractTransactionsOutcomeParser.spec.ts
@@ -93,7 +93,7 @@ describe("test smart contract transactions outcome parser", () => {
         const transactionOnNetwork = new TransactionOnNetwork({
             nonce: 7n,
             function: "getUltimateAnswer",
-            smartContractResults: [new SmartContractResult({ data: Buffer.from("@6f6b@2a") })],
+            smartContractResults: [new SmartContractResult({ data: Buffer.from("QDZmNmJAMmE=", "base64") })],
         });
 
         const parsed = parser.parseExecute({ transactionOnNetwork });
--- a/src/smartContracts/smartContractTransactionsOutcomeParser.ts
+++ b/src/smartContracts/smartContractTransactionsOutcomeParser.ts
@@ -142,7 +142,7 @@ export class SmartContractTransactionsOutcomeParser {
         const eligibleResults: SmartContractResult[] = [];
 
         for (const result of transactionOnNetwork.smartContractResults) {
-            const matchesCriteriaOnData = result.data.toString().startsWith(ARGUMENTS_SEPARATOR);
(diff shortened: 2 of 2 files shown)


src/smartContracts/smartContractTransactionsOutcomeParser.ts
```diff
@@ -142,7 +142,7 @@ export class SmartContractTransactionsOutcomeParser {
         const eligibleResults: SmartContractResult[] = [];
 
         for (const result of transactionOnNetwork.smartContractResults) {
-            const matchesCriteriaOnData = result.data.toString().startsWith(ARGUMENTS_SEPARATOR);
+            const matchesCriteriaOnData = Buffer.from(result.data).toString("utf-8").startsWith(ARGUMENTS_SEPARATOR);
```
andreibancioiu, line 145: Good point, applying `toString()` on `data: Uint8Array` was tricky.

https://github.com/multiversx/mx-sdk-js-core/blob/main/src/transactionsOutcomeParsers/resources.ts#L10

src/smartContracts/smartContractTransactionsOutcomeParser.spec.ts
```diff
@@ -93,7 +93,7 @@ describe("test smart contract transactions outcome parser", () => {
         const transactionOnNetwork = new TransactionOnNetwork({
             nonce: 7n,
             function: "getUltimateAnswer",
-            smartContractResults: [new SmartContractResult({ data: Buffer.from("@6f6b@2a") })],
+            smartContractResults: [new SmartContractResult({ data: Buffer.from("QDZmNmJAMmE=", "base64") })],
```
andreibancioiu, line 96: I think the inconsistencies between API and Proxy are not yet reconciled at the SDK level, thus a further fix is needed.

 - https://github.com/multiversx/mx-sdk-js-core/blob/main/src/core/transactionOnNetwork.ts#L71
 - https://github.com/multiversx/mx-sdk-js-core/blob/main/src/core/transactionOnNetwork.ts#L141

On proxy, it's an UTF-8 string (it's sub-components being hex-encoded):
 - https://devnet-gateway.multiversx.com/transaction/5647414afcd17d74ba0223837cefa47347946d806e1196d3e16558c881fc955d?withResults=true

On API, it's a base64-encoded string:
 - https://devnet-api.multiversx.com/transactions/5647414afcd17d74ba0223837cefa47347946d806e1196d3e16558c881fc955d


Good catch :tada: 

We'll discuss this a bit internally, then come back.

andreibancioiu: Actually, I think we can approve this PR, since there's no change in the implementation with respect to this issue - there's only the fix  in `smartContractTransactionsOutcomeParser.ts`. And we can handle the Proxy - API inconsistency in a separate PR.

andreibancioiu (approved):

danielailie (approved):

PR #607, opened 2025-05-06
Update multisig controller and factory to use composition and not inheritance

danielailie:


andreibancioiu (dismissed):

src/multisig/multisigTransactionsFactory.ts
```diff
@@ -411,4 +415,21 @@ export class MultisigTransactionsFactory {
             arguments: [VariadicValue.fromItems(...actionIdsArgs)],
         });
     }
+
+    protected argsToTypedValues(args: any[], endpoint?: EndpointDefinition): TypedValue[] {
```
popenta, line 419: maybe we should also allow bytes arguments?

popenta (approved):

andreibancioiu (approved):

PR #609, opened 2025-05-07
Delete unused files

danielailie:


popenta (approved):
Even though these were not used anymore it's still a breaking change. We should at least mention this in the release notes.

andreibancioiu (approved):

PR #611, opened 2025-05-09
Add get transactions for address for api network provider

popenta:


src/networkProviders/apiNetworkProvider.ts
```diff
@@ -149,6 +149,12 @@ export class ApiNetworkProvider implements INetworkProvider {
         return transaction;
     }
 
+    async getTransactions(address: Address): Promise<TransactionOnNetwork[]> {
```
andreibancioiu, line 152: As opposed to the python implementation, here we don't support options? Shouldn't we?

andreibancioiu (approved):

mradian1 (approved):

src/networkProviders/apiNetworkProvider.ts
popenta: We can add it in the TODO list. If someone needs to do a custom request, they can use `doGetGeneric()`.

PR #626, opened 2025-06-29
Improve network providers address hrp compatibility

michavie:
Previously was not compatible with Sovereign Chains that use different HRP than `erd`.


src/networkProviders/accounts.ts
```diff
@@ -50,7 +52,7 @@ export class AccountOnNetwork {
     static fromProxyHttpResponse(payload: any): AccountOnNetwork {
         const result = new AccountOnNetwork();
 
-        result.address = payload["address"] ? new Address(payload["address"]) : Address.empty();
+        result.address = payload["address"] ? Address.newf(payload["address"]) : Address.empty();
```
andreibancioiu, line 55: Typo / `newFromBech32`

PR #652, opened 2025-08-05
Allow passing network options from Entrypoint to Network Provider

BubuMVX:
Hello,

### Summary

This PR adds support for passing custom network options from the `Entrypoint` to the `Network Provider`. This allows developers to configure headers and other request-related parameters when initializing the entrypoint.

### Changes

- Added a new `networkProviderOptions` field in the Entrypoint options.
- Ensured `clientName` can be set via either `options.clientName` or fallback to `User-Agent` header.

### Motivation

Provides more flexibility for advanced developers needing to customize the network layer (e.g. for authentication, logging, or custom headers).

### DIFF
--- a/src/entrypoints/entrypoints.ts
+++ b/src/entrypoints/entrypoints.ts
@@ -16,7 +16,7 @@ import { GasLimitEstimator } from "../gasEstimator";
 import { GovernanceController, GovernanceTransactionsFactory } from "../governance";
 import { MultisigTransactionsFactory } from "../multisig";
 import { MultisigController } from "../multisig/multisigController";
-import { ApiNetworkProvider, ProxyNetworkProvider } from "../networkProviders";
+import { ApiNetworkProvider, NetworkProviderConfig, ProxyNetworkProvider } from "../networkProviders";
 import { INetworkProvider } from "../networkProviders/interface";
 import { SmartContractTransactionsFactory } from "../smartContracts";
 import { SmartContractController } from "../smartContracts/smartContractController";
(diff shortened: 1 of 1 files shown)


src/entrypoints/entrypoints.ts
```diff
@@ -34,19 +34,24 @@ export class NetworkEntrypoint {
...
+        networkProviderOptions?: NetworkProviderConfig;
         chainId: string;
         clientName?: string;
         withGasLimitEstimator?: boolean;
         gasLimitMultiplier?: number;
     }) {
+        options.networkProviderOptions = options.networkProviderOptions || {};
+
+        if (options.clientName) {
+            options.networkProviderOptions.clientName = options.clientName;
```
andreibancioiu, line 46: Such a line:

```
options.networkProviderOptions.[...] = ...
```

Will mutate the input argument `networkProviderOptions`, which might cause downstream issues (if that input argument is re-used by the caller). Maybe clone the options (below is just an example, not an actual suggestion)?

```
networkProviderOptions = { ...options.networkProviderOptions, ...{ clientName: options.clientName } };
```

Maybe a cleaner way would be to simply drop the input argument `clientName`, and only have `networkProviderConfig`? However, that would be a breaking change, and we shouldn't apply it (yet!).

src/entrypoints/entrypoints.ts
```diff
@@ -34,19 +34,24 @@ export class NetworkEntrypoint {
     constructor(options: {
         networkProviderUrl: string;
         networkProviderKind: string;
+        networkProviderOptions?: NetworkProviderConfig;
```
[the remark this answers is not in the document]
andreibancioiu: In the constructors of `ProxyNetworkProvider` and `ApiNetworkProvider`, this is named "config". Maybe, for consistency, rename the parameter to `networkProviderConfig`?

src/entrypoints/entrypoints.ts
```diff
@@ -34,19 +34,24 @@ export class NetworkEntrypoint {
...
         clientName?: string;
         withGasLimitEstimator?: boolean;
         gasLimitMultiplier?: number;
     }) {
+        options.networkProviderOptions = options.networkProviderOptions || {};
+
+        if (options.clientName) {
+            options.networkProviderOptions.clientName = options.clientName;
+        } else if (options.networkProviderOptions?.headers?.["User-Agent"]) {
+            options.networkProviderOptions.clientName = options.networkProviderOptions.headers["User-Agent"];
```
andreibancioiu, line 48: This is an opaque trick (a hack), and assumes that "client name" and "header user agent" are synonyms for network providers (somehow, it breaks a bit their encapsulation). 

Maybe drop this logic, and, if desired, a separate PR can be made, but not for the entrypoints. See the file below for a possible place of adjustment:
 - https://github.com/multiversx/mx-sdk-js-core/blob/main/src/networkProviders/userAgent.ts

src/entrypoints/entrypoints.ts
BubuMVX: About the `clientName` and `User-Agent` header, my opinion is:

- `clientName` is good for general usage, as you want to collect stats. It's easy to set and abstracts the technical implications. And you already build a custom User-Agent from that, that looks like a good approach.
- More advanced users may want to force a very specific `User-Agent`. This shouldn't be altered later in the Network Provider.
- One of them must be specified for sure, otherwise you get the console error message and the "unknown" fallback. With a priority to be discussed (probably User-Agent > clientName).

andreibancioiu (approved):

PR #657, opened 2025-09-15
Add validator pem and extend validator signer

danielailie:


popenta (commented):
looks good, but add unit tests

src/wallet/validatorSigner.ts
```diff
@@ -6,15 +6,19 @@ import { ValidatorPEM } from "./validatorPem";
...
+    private readonly secretKey: ValidatorSecretKey;
 
     constructor(secretKey: ValidatorSecretKey) {
         this.secretKey = secretKey;
     }
     /**
      * Signs a message.
      */
-    async signUsingPem(pemText: string, pemIndex: number = 0, signable: Buffer | Uint8Array): Promise<Uint8Array> {
+    static async signUsingPem(
```
andreibancioiu, line 17: Tiny breaking change in theory, but can be in minor version - since, actually, it's a "design correction change" (it's a fix).

andreibancioiu (approved):

andreibancioiu (approved):

src/wallet/validatorPem.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+            fs.unlinkSync(savedPath); // cleanup
+        }
+    });
+
+    it("should save pem from file", async function () {
+        const contentExpected = fs.readFileSync(pemPath, "utf-8").trim();
+
+        // fromFile is async → await
+        const pem = await ValidatorPEM.fromFile(pemPath);
+        pem.save(savedPath);
```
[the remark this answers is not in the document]
danielailie: I added tests for toText, save is already tested via the other tests

src/wallet/validatorSigner.ts
popenta: as discussed with @andreibancioiu, I think we can mark this as deprecated since we already have the sign method and this does not seem to bring significant benefits.

popenta (approved):

